PR #2, opened 2025-02-27
Pbi 1 convert jet uml class diagram files to code

KenKomKom:
Implemented functionality to parse a class diagram from a JSON string, extracting class names, attributes, and relationships. This allows the translation from JSON to premade classes in the application

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,7 +1,25 @@
+from fastapi.responses import FileResponse
 from fastapi import FastAPI
+from starlette.background import BackgroundTasks
+from app.utils import remove_file
+from app.model import DownloadRequest
 
 app = FastAPI()
 
 @app.get("/")
 def read_root():
     return {"message": "Hello, FastAPI World!"}
+
+@app.post("/download/")
+async def download_file(request : DownloadRequest, background_tasks: BackgroundTasks):
+    file = "a.py"
+    
+    # Create a temporary file for demonstration
+    with open(file, "w") as f:
+        f.write(request.content)
+
+    response = FileResponse(path=file, filename=file, media_type='file')
+
+    background_tasks.add_task(remove_file, file)
+
+    return response
--- a/tests/test_parse_class.py
+++ b/tests/test_parse_class.py
@@ -185,6 +185,22 @@ class TestParseJsonToObject(unittest.TestCase):
         ro = class_from_id.add_relationship.call_args[0][0]  # Extract the relationship object
         self.assertIsInstance(ro, OneToOneRelationshipObject)
 
+    def test_parse_relationships_inheritance(self):
+        uml_json = {'edges':[{
+                'start': 1,
+                'end': 2,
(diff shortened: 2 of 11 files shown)


DaWanAnOnli:
- Next time, just make sure that commits with the 'implement' message will actually pass all tests. As from here, we can't tell which tests are passed for each commit. Therefore if an 'implement' commit is still [RED], we don't know whether the failed tests are related to the  tests this commit is trying to pass or not.

app/main.py
```diff
@@ -1,7 +1,31 @@
...
+async def download_file(request : DownloadRequest, background_tasks: BackgroundTasks):
+    file = request.filename+".py"
+
+    if '/' in request.filename or '\\' in request.filename:
+        raise HTTPException(status_code=400, detail="/ not allowed")
+
+    if os.path.exists(file):
+        raise HTTPException(status_code=400, detail="The file already exists.")
+
+    with open(file, "w") as f:
```
MightyZanark, line 24: for opening files in async functions, use `aiofiles` instead so it does not block the thread, might also want to use the `tempfile` module instead of creating and deleting it manually

MightyZanark:
Other than the ones commented, might want to use the linter and formatter first as well

tests/test_parse_class.py
```diff
@@ -0,0 +1,244 @@
...
+                "type": "ClassNode"
+                }
+            ]
+        }"""
+        parser = ParseJsonToObject(data)
+        result = parser.parse_classes()
+        self.assertEqual(result, parser._ParseJsonToObject__classes)
+        self.assertEqual(parser._ParseJsonToObject__json, json.loads(data))
+        self.assertEqual(len(parser._ParseJsonToObject__json["nodes"]), 1)
+        self.assertEqual(parser._ParseJsonToObject__json["nodes"][0]["name"], "Pengelola")
```
NtapSlur, line 102: You can add test to verify what the program will do if it detects a missing class name since it can cause an error if it's proccessed

NtapSlur: Here are some other notes that can be taken:
- It is better to change the name of the file **parse_json_to_object.py** since there will also be a parser for the Sequence Diagram. It will be hard to identify which parser is for which diagram if it is not specified in the file name for the UML Diagram type. 
- It is also better if the test case includes the other edge case of the program like duplicate class name or attribute

Other than that, the rest is already good! 👍

app/main.py
KenKomKom: By aiofiles, you mean these right? [aiofiles](https://pypi.org/project/aiofiles/) [https://docs.python.org/3/library/tempfile.html](https://docs.python.org/3/library/tempfile.html).
I'll fix it for the next pull request

tests/test_parse_class.py
KingOnyxx: Thanks for noticing! I'll try to implement it.

PR #3, opened 2025-02-27
Add linter, formatter, and pre-commit hooks to project

MightyZanark:
- Added the `ruff` linter and formatter to make sure our code style is consistent.
- Added pre-commit hooks that checks our code style before we can make a commit. If there is any violation, then we can't make a commit until it is fixed.
- Added pre-commit hooks that checks our files to check if there is any accidental private credentials added to the staging area before we can make a commit. If there is, we need to remove it before being able to commit.
- Edited `ci.yml` so it runs the test on **every** pull requests instead of only ones targeting `main`, `staging`, or `dev`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,14 +6,14 @@ on:
       - main
       - staging
       - dev
+    paths:
+      - '**.py'
   pull_request:
-    branches:
-      - main
-      - staging
-      - dev
+    paths:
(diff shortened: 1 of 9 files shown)


KenKomKom:
Looks good to me. cheers

NtapSlur:
It's already good for the commit! And it's nice using Linter as source code analyzer!

PR #5, opened 2025-03-03
Pbi 2 8 write views functions in views.py

DaWanAnOnli:
Finished task PBI-2-8

- create methods in intermediate classes to convert instances of ControllerMethodObject into django code.
- create ModelsElement and ViewsElement as containers for all elements contained in the to-be-generated models.py and views.py files respectively
- The return values are still in string, they are not yet put into any file format
- The conversion can handle methods with or without arguments, with or without return values (hence return variables), and with or without condition (if clause).

### DIFF
--- a/app/element_objects.py
+++ b/app/element_objects.py
@@ -230,6 +230,20 @@ class ControllerMethodObject(AbstractMethodObject):
     def add_call(self, call_object : AbstractMethodCallObject):
         self.__calls.append(call_object)
 
+    def print_django_style(self):
+        if not self.get_name():
+            raise TypeError("method cannot be empty")
+        result = StringIO()
+        result.write(f"def {self.get_name()}(request")
+        for parameter in self.get_parameters():
+            result.write(f", {parameter.get_name()}")
+        result.write("):\n\t")
+        for abstract_method_call_object in self.__calls:
+            result.write(abstract_method_call_object.print_django_style())
+            result.write("\n\t")
+        result.write("\n")
+        return result.getvalue()
+
 class ClassMethodCallObject(AbstractMethodCallObject):
     def __init__(self):
         super().__init__()
--- a/app/element_objects.py
+++ b/app/element_objects.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
-from abc import ABC, abstractmethod
+
+from abc import ABC
 from io import StringIO
-
-
-
 
 
 class ClassObject:
@@ -73,10 +71,10 @@ class AbstractMethodObject(ABC):
     def set_returnType(self, returnType: TypeObject):
         self.__returnType = returnType
 
-    def get_name(self):
+    def get_name(self) -> str:
         return self.__name
-    
-    def get_parameters(self):
+
+    def get_parameters(self) -> list[ParameterObject]:
         return self.__parameters
 
 
@@ -89,6 +87,7 @@ class ClassMethodObject(AbstractMethodObject):
         if class_method_call is None:
             raise Exception("Cannot add None to ClassMethodCallObject!")
         self.__calls.append(class_method_call)
+
 
 class AbstractRelationshipObject(ABC):
     def __init__(self):
@@ -128,8 +127,9 @@ class ParameterObject:
     def set_type(self, type: str):
         self.__type = type
 
-    def get_name(self):
+    def get_name(self) -> str:
         return self.__name
+
 
 class AbstractMethodCallObject(ABC):
     def __init__(self):
@@ -153,10 +153,10 @@ class AbstractMethodCallObject(ABC):
     def set_returnVarName(self, returnVarName: str):
         self.__returnVarName = returnVarName
 
-    def set_condition(self, condition):
+    def set_condition(self, condition: str):
         self.__condition = condition
 
-    def print_django_style(self):
+    def print_django_style(self) -> str:
         result = StringIO()
         print("condition:", self.__condition)
         if self.__condition:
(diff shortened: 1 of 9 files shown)


app/element_objects.py
```diff
@@ -183,25 +218,77 @@ def __init__(self):
...
+
+    def set_caller(self, method_object: ClassMethodObject):
+        if method_object is None:
             raise Exception("ClassMethodObject cannot be SET to be None!")
-        self.__caller = method_object
\ No newline at end of file
+        self.__caller = method_object
+
+
+class FileElements(ABC):
```
MightyZanark, line 261: What might be the benefit of this class and its subclasses for our project? Is it just a wrapper for what can be converted to code?

app/element_objects.py
```diff
@@ -183,25 +218,77 @@ def __init__(self):
...
+class ViewsElements(FileElements):
+    def __init__(self, file_name: str):
+        super().__init__(file_name)
+        self.__class_methods: list[ClassMethodObject] = []
+        self.__controller_methods: list[ControllerMethodObject] = []
+
+    def print_django_style(self) -> str:
+        result = StringIO()
+
+        for class_method_object in self.__class_methods:
```
MightyZanark, line 283: Why does this loop exists if it does nothing?

app/element_objects.py
```diff
@@ -183,25 +218,77 @@ def __init__(self):
...
+            pass
+
+        for controller_method_object in self.__controller_methods:
+            result.write(controller_method_object.print_django_style())
+        return result.getvalue()
+
+    def add__class_method(self, class_method_object: ClassMethodObject):
```
MightyZanark, line 290: For the method names, I think its better if its just `add_class_method` and `add_controller_method` instead of the double underscores in the start. The people using this method should not need to know what the variable inside this class is called, they just need to know that this method does what it is expected to do.

app/element_objects.py
DaWanAnOnli: It's a place to store all intermediate class instances made after parsing. To make sure that these instances are parts of the same to-be-generated file, they need to be stored in a container class. This container class has a method to generate django code, i.e print_django_style method.

Next time I plan to make print_django_style method as an abstract method in FileElements class, so that ModelsElements and ViewsElements must implement the method.

app/element_objects.py
```diff
@@ -1,175 +1,210 @@
...
-class ArgumentObject():
+    def set_return_var_name(self, return_var_name: str):
+        self.__return_var_name = return_var_name
+
+    def set_condition(self, condition: str):
+        self.__condition = condition
+
+    def print_django_style(self) -> str:
+        result = StringIO()
+        print("condition:", self.__condition)
```
MightyZanark, line 162: You should ideally not have print statements outside of debugging context

app/element_objects.py
DaWanAnOnli: This part of the code is reserved for different tasks, i.e PBI-1-7 and PBI-2-9.  Tha'ts why it's still empty. 

I will clarify this part by adding //TODO comment and the tasks that correspond to this line.

app/element_objects.py
DaWanAnOnli: Noted. This is something I overlooked. Thanks for the feedback

app/element_objects.py
DaWanAnOnli: Noted. I overlooked this part. Thanks for the feedback

MightyZanark:
Other than the ones I specifically commented on, some other nitpicks are:
- Might want to use spaces instead of tabs as recommended by [PEP8](https://peps.python.org/pep-0008/#tabs-or-spaces).
- If a string isn't supposed to be a multiline string, you can just use single double quotes instead of triple.

app/element_objects.py
MightyZanark: I see. Yeah it is better to add a TODO comment or similar if it is going to be implemented later. Resolving this comment

app/element_objects.py
MightyZanark: Noted. So these classes will probably shine when we get to the stage where there are multiple files we need to convert, correct?

DaWanAnOnli: Correct. We must make sure that multiple sequence diagram files only correspond to one ViewsElements object (as they won't affect models.py file, thus not needing to correspond to ModelsElements object)

KenKomKom:
I want to ask on the reason why Optional is used, wouldn't 
variable : Class = None do the same thing, or am I mistaken?

app/element_objects.py
```diff
@@ -1,175 +1,208 @@
...
     def __init__(self):
-        self.__name : str = ""
-        self.__parameters : list[ParameterObject] = []
-        self.__returnType : TypeObject = None
-
-    def __str__(self):
-        return f'''MethodObject:\n\tname: {self.__name}\n\tparameters: {self.__parameters}\n\treturnType: {self.__returnType}'''
-    
-    def set_name(self, name):
+        self.__name: str = ""
```
NtapSlur, line 56: Why isnt't the parameter name assigned to the attribute name?

app/element_objects.py
```diff
@@ -1,175 +1,208 @@
...
+        self.__return_type: Optional[TypeObject] = None
+
+    def __str__(self) -> str:
+        return (
+            f"MethodObject:\n\tname: {self.__name}\n\tparameters: {self.__parameters}"
+            f"\n\treturn_type: {self.__return_type}"
+        )
+
+    def set_name(self, name: str):
         self.__name = name
```
NtapSlur, line 67: Oops, solved in this line of code!

NtapSlur:
For me, the code is already good! You even used ABC for the Abstract Class which is the best practice!

PR #10, opened 2025-03-03
Pbi 1 11 generate downloadable files

KenKomKom:
Finished PBI 1-11

In this pull request, I have done the following:
create an API for convertion that receives a JSON in the request body that has keys "filename" and "content" both of which are String.
parsed said request into classes
writes said classes into files and sends it to the user as models.py 

Note:
The /convert API cannot handle parsing for sequence diagram


Not0nline:
can you make a readme for the format for the template of the JSON body that is needed for the API that will be accepted? thanks
- Adrial

DaWanAnOnli:
There are two feedbacks for this pull request.

Single Responsibility Principle:
the methods parse_classes, parse_relationships, and __validate_amount are very long. Is there a way to break these methods into smaller ones? This will help with readability and separation of concerns.
(P.S. I'm pretty sure I've commented on this issue on a previous pull request. But until now it is still not fixed Please fix this to avoid complications in the long run)
Documentation:
Add docstring for each method, preferably with examples. So we know what input and expected output of the method.
Keep up the good work!

app/main.py
```diff
@@ -1,32 +1,71 @@
...
+    response = FileResponse(path=file, filename=file, media_type="file")
 
     background_tasks.add_task(remove_file, file)
 
-    return response
\ No newline at end of file
+    return response
+
+
+def remove_file(path: str) -> None:
```
MightyZanark, line 42: I think its better to keep this in `utils.py` as before.

app/main.py
```diff
@@ -1,32 +1,71 @@
...
+    parser.parse_relationships(classes)
+
+    # A faster way to build string
+    # https://stackoverflow.com/questions/2414667/python-string-class-like-stringbuilder-in-c
+    response_content = "".join(
+        [model_class.to_models_code() for model_class in classes]
+    )
+    response_content = response_content.strip()
+    response_content += "\n" if len(response_content) != 0 else ""
+    async with httpx.AsyncClient(base_url=str(fastapi_request.base_url)) as client:
```
MightyZanark, line 63: This honestly seems a bit too convoluted. Why don't the convert endpoint also make the file and send it back in the response? Making it send requests to another endpoint just to download seems like extraneous HTTP request and API call.

app/parse_json_to_object.py
```diff
@@ -19,6 +19,8 @@ def parse_classes(self):
                 raise Exception("Error: class not found in the json")
             a = ClassObject()
             a.set_name(classes['name'])
+            print(a.get_name())
```
MightyZanark, line 22: Should remove print statements outside of debugging contexts

MightyZanark (commented):
Overall looks good, just want some clarification on the `/convert` endpoint as commented.

app/main.py
KenKomKom: I don't know whether there needs to be other actions that needs the /convert url so I thought that it will be better to split them both. The reason that /download needs to be an API of its own is because it needs background task to delete the file which it cannot have just by being a function. But if there isn't any, we can just join the /download into /convert

app/parse_json_to_object_class.py
```diff
@@ -0,0 +1,262 @@
...
+                                    class_obj.add_field(attr)
+                                else:
+                                    raise ValueError(
+                                        "Error: attribute name or type is not valid"
+                                    )
+
+            self.__classes.append(class_obj)
+        return self.__classes
+
+    def parse_relationships(self, classes):  # noqa: ANN001, ANN201, ANN202,
```
KingOnyxx, line 146: I think you should remove this comment, so ruff can check your code quality

app/main.py
MightyZanark: I guess we'll have to see later whether we should combine in or not. In regards to creating and deleting the file, I think we should find if we can just send the raw bytes instead. As far as I know, we can use the `Content-Disposition` header to signal that we are sending an attachment with some filename.

KenKomKom (commented):
I changed all set_returnType into set_return_type

PR #13, opened 2025-03-03
Pbi 2 convert jet uml sequence  diagram files to code

DaWanAnOnli:
Finished task PBI-2-8

- create methods in intermediate classes to convert instances of ControllerMethodObject into django code.
- create ModelsElement and ViewsElement as containers for all elements contained in the to-be-generated models.py and views.py files respectively
- The return values are still in string, they are not yet put into any file format
- The conversion can handle methods with or without arguments, with or without return values (hence return variables), and with or without condition (if clause).
- Configure sonarcloud to run on pull requests


Not0nline:
hello, after reviewing your code, maybe you can add comments to each method for a short description on what they do
thanks

KenKomKom:
From what I understand, the code already seems pretty robust. But I don't think it has handled the case where the parameter is a type of List, and I think that will be necessary for further development. You can probably use regex for this, and you can see an example of the regex in parse_json_object_class.

KingOnyxx:
Thanks for setting up the cosmic-ray! You can add documentation to explain what you implemented, but other than that, i think your code is great.

MightyZanark:
Looks good! Although I think it is better to add stuff like `cosmic-ray` outside of feature branches, so that we don't need to wait for the feature branch to be merged to use those tools. Adding a tool to help our development shouldn't be gated behind us having to merge a feature branch that may or may not be finalized yet.

DaWanAnOnli:
> hello, after reviewing your code, maybe you can add comments to each method for a short description on what they do thanks

Okay I will do that next time as it is currently not the most critical task


> From what I understand, the code already seems pretty robust. But I don't...

From the start I never plan to have type casting for parameters. So I believe this will not be necessary. However this is something that we must discuss further in person.

> Thanks for setting up the cosmic-ray! You can add documentation to explain what you implemented, but other than that, i think your code is great.

No worries.

> Looks good! Although I think it is better to add stuff like cosmic-ray outside of feature branches, so that we ...
Noted. For future tools I will implement this principle.

PR #14, opened 2025-03-03
100% code coverage for parse_classes

KingOnyxx:
This pull request includes several changes to the `app/parse_json_to_object_class.py` file to improve error handling and parsing logic, as well as the addition of new tests in `tests/test_parse_class.py` to ensure the robustness of the parsing functionality.

### DIFF
--- a/tests/test_parse_class.py
+++ b/tests/test_parse_class.py
@@ -65,6 +65,15 @@ class TestParseJsonToObjectClass(unittest.TestCase):
             parser.parse_classes()
             self.assertEqual(context.exception, "Error: nodes not found in the json")
 
+    def test_missing_nodes(self):
+        data = """{
+            "diagram": "ClassDiagram"
+        }"""
+        parser = ParseJsonToObjectClass(data)
+        with self.assertRaises(ValueError) as context:
+            parser.parse_classes()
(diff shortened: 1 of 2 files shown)


KenKomKom:
I think for the most part, your work is good, but I think for further development we need a way to validate duplicate methods and attributes

DaWanAnOnli:
I think you did a good job in making sure your code concur with Python's best practice. 
image
In this case, obviously the best way to check whether "nodes" is in data.keys is by using "in" instead of == None.

Now, you should take the time to make documentation to your methods by adding docstrings. Make sure for each docstring you add some input-output pairs as examples to ease other programmers who are reading your code.

tests/test_parse_class.py
```diff
@@ -127,6 +136,64 @@ def test_missing_attributes(self):
             parser._ParseJsonToObjectClass__json["nodes"][0]["type"], "ClassNode"
         )
 
+    def test_missing_method_name(self):
+        data = """{
```
MightyZanark, line 140: Might want to move these data to a file instead of having it in text here.

MightyZanark (commented):
Looks good, just some small comments that you might want to change before merging.

PR #17, opened 2025-03-04
PBI 1-8: Write Methods to views.py

MightyZanark:
Finished task PBI 1-8. This pull request is the second part of #4 
- Added `to_views_code` to `ClassObject`
- Fixed bad test where there were some hardcoded values that I forgot to change
- Made the `AbstractMethodObject` immutable by making its getter return a copy instead of a reference to the real value to reduce accidental change. This change have passed the current tests on `AbstractMethodObject`.


app/models/diagram.py
```diff
@@ -25,6 +27,22 @@ def to_models_code(self) -> str:
...
+            res = f"from .models import {self.__name}\n"
+
+        if self.__parent is not None:
+            res += self.__parent.to_views_code()
+
+        for method in self.__methods:
+            res += method.to_views_code()
+
+        return res
+
```
KenKomKom, line 45: I suggest making this in bottom-up approach just to be safe from stack overflow error, like
parent = self
while parent != None:
    res+=f"from .models import {parent.__name}\n"
    parent = parent.parent
parent = self
while parent != None:
    for method in parent.get_methods:
            res += method.to_views_code()
    parent = parent.parent

but if you don't want to, I can understand why, it really is a nice and simple code.

MightyZanark: I did think about the possibility of a stack overflow error, but I thought that this way of doing things is also still temporary as I am thinking of moving it to Jinja2 soon.

PR #18, opened 2025-03-05
PBI 2 - Create parser for ControllerMethodObject (views)

NtapSlur:
Finished PBI 2-4:
- Added method parse() that can parse all views method
- Testcase given:
1. Positive: Valid JSON to be parsed and assert all expected views method
2. Negative: Handled in previous PR about validate JSON structure
3. Edge: Duplicate class name, attribute, and method


MightyZanark:
Other than the ones given by Gemini, I think you should throw specific exceptions instead of a generic `Exception`. This makes the exception more informative and makes us not need to catch every exception (because in Python, most exceptions, even custom ones, are instances of the base `Exception`, so doing `except Exception` will catch a lot more exception than you might think).

Otherwise, looks good!

app/parse_json_to_object_seq.py
```diff
@@ -71,3 +80,110 @@ def validate_json(self, data: object) -> bool:
...
+
+                else:
+                    # TODO: Create for Class Object
+                    continue
+
+                method_label = edge["label"].split(" ")
+                is_name_setup = False
+                duplicate_attribute_checker = dict()
+
+                for value in method_label:
```
KingOnyxx, line 161: need documentation for this for-loop block

tests/test_parse_json_to_object_seq.py
```diff
@@ -461,3 +461,1727 @@ def test_negative_set_empty_json(self):
...
+    },
+    {
+      "middleLabel": "hasTanggunganResult",
+      "start": 28,
+      "end": 9,
+      "type": "ReturnEdge"
+    }
+  ],
+  "version": "3.8"
+}
```
KingOnyxx, line 878: i think you can put these json inputs in a new file so the test can be reviewed better.

KingOnyxx (commented):
I have reviewed your code and i want to give you a couple of feedbacks.

tests/test_parse_json_to_object_seq.py
KenKomKom: Yes I absolutely agree with @KingOnyxx's comment on this. It is absolutely disgusting that the input is 500 line long just for a 5 line test. It needs to be moved to a new file ASAP. then in the test read it as an input

app/parse_json_to_object_seq.py
```diff
@@ -71,3 +80,110 @@ def validate_json(self, data: object) -> bool:
 
         except Exception:
             return False
+
+    def get_controller_method(self) -> list[ControllerMethodObject]:
+        return self.__controller_method
+
+    def parse(self):
```
DaWanAnOnli, line 87: For me this method is too long. It can be divided into smaller method by, for example, by extracting each for loop as its own method. This conforms to the Single Responsibility Principle. It makes the code more modular and readable. Otherwise, Sonarqube will also have a complaint about this method for being too complex

app/parse_json_to_object_seq.py
```diff
@@ -2,10 +2,19 @@
...
+
 
 class ParseJsonToObjectSeq:
     def __init__(self):
         self.__json = None
+        self.__class_object: dict = dict()
+        self.__controller_method: list[ControllerMethodObject] = []
+        self.__call_nodes: dict = dict()
+        self.__edges: list = []
+        self.__implicit_parameter_nodes = dict()
```
DaWanAnOnli, line 17: You can add more specific type casting for all the dict objects, in particular the type for the key and value for each dict. so that we know how the mapping works for the dictionary. Because now we need to find an example of the usage of the dict to determine the types of the keys and the values. By having a more specific type casting, the code will be more readable and less tiring for other programmers trying to figure out what to do.

PR #21, opened 2025-03-07
Pbi 3 1 modify api

KenKomKom:
In this pull request, I have changed main.py /convert to take multiple files which are UML class diagrams. The example of such request is attached in the form of a json for postman.
Happy weekends, cheers!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -44,54 +44,53 @@ async def convert(
             status_code=400, detail="number of Filename and Content is incosistent"
         )
 
-    for file_name, content in zip(request.filename, request.content):
-        writer = ModelsElements(file_name)
-        json_content = content[0]
+    response_content_models = ""
+    response_content_views = ""
 
+    for file_name, content in zip(request.filename, request.content):
+        json_content = content[0]
         if isinstance(json_content, str):
             json_content = json.loads(json_content)
 
-        # Only parse and writes to models.py if from class diagram
         if (
             json_content["diagram"] is not None
             and json_content["diagram"] == "ClassDiagram"
         ):
-            classes = writer.parse(json_content)
-            response_content = writer.print_django_style()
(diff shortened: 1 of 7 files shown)


app/main.py
```diff
@@ -35,43 +36,67 @@ async def download_file(request: DownloadRequest) -> FileResponse:
...
+                request=DownloadRequest(
+                    filename=file_name, content=response_content, type="_models"
+                ),
+            )
+
+            writer = ViewsElements(file_name)
+            for model_class in classes:
+                for method in model_class.get_methods():
+                    writer.add_class_method(method)
+            response_content = writer.print_django_style()
```
DaWanAnOnli, line 72: I see here that you parse class diagram, write to models.py, make models.py downloadable, and then you parse sequence diagram, write to views.py, and then make views.py downloadable.

This is not right, as class diagrams will affect both models.py and views.py. Sequence diagrams will also affect both models.py and views.py. 

I recommend that ModelsElements and ViewsElements are both intialized at the start, as the following processes will modify both. Then, you make both of them downloadable at the very end, not one before the other.

KenKomKom: ok I'll fix it now

app/models/elements.py
```diff
@@ -23,12 +23,20 @@ def __init__(self, file_name: str):
         super().__init__(file_name)
         self.__classes: list[ClassObject] = []
 
+    """
+    Parses ClassDiagram to classes
+    """
```
MightyZanark, line 28: Docstrings are supposed to be *inside* of the function/class they are documenting, not outside. For an example, you can see [here](https://github.com/fastapi/fastapi/blob/643d2845de09a437b955a163ab25eb0887bfd0e1/fastapi/datastructures.py#L133).

PR #22, opened 2025-03-08
PBI 2-6: Parse self calls on sequence diagram

MightyZanark:
Finished implementing parsing of self calls in sequence diagram. Here is the summary:
- Added self call parsing as part of the regular parsing workflow
- Refactored the for loop to generate a `ClassMethodObject` to be sequential and uses regular expressions instead
- Added call tree depth check to counter act possible DoS scenarios. Currently the amount of self call depth allowed is 5, but can be easily changed by changing the variable value.

If there are any edge cases or even regular cases that I might've missed, please do tell me


app/parse_json_to_object_seq.py
```diff
@@ -0,0 +1,276 @@
...
+                call_obj.set_return_var_name(ret_var)
+            caller_method.add_class_method_call(call_obj)
+
+    def check_call_depth(self, rev_call_tree: dict[int, int], callee: int) -> int:
+        depth = 0
+        caller = rev_call_tree[callee]
+        while caller > 0:
+            caller = rev_call_tree.get(caller, -1)
+            depth += 1
+        return depth
```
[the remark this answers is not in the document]
MightyZanark: Thanks for the catch, implemented

app/models/diagram.py
```diff
@@ -0,0 +1,157 @@
...
+class AbstractRelationshipObject(ABC):
+    def __init__(self):
+        self.__source_class: Optional[ClassObject] = None
+        self.__target_class: Optional[ClassObject] = None
+        self.__sourceClassOwnAmount: str = ""
+        self.__targetClassOwnAmount: str = ""
+
+    def set_source_class(self, source_class: ClassObject):
+        if source_class is None:
+            raise Exception("Source Class cannot be SET to be None!")
```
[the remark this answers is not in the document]
MightyZanark: Resolving as this is not part of my code

app/models/diagram.py
```diff
@@ -25,6 +27,22 @@ def to_models_code(self) -> str:
...
+        res = ""
+        if len(self.__methods) > 0:
+            res = f"from .models import {self.__name}\n"
+
+        if self.__parent is not None:
+            res += self.__parent.to_views_code()
+
+        for method in self.__methods:
+            res += method.to_views_code()
+
```
DaWanAnOnli, line 43: Again, I would suggest using StringIO for this instead of string concatenation as Python strings are immutable. So by concatenating, you're actually creating a new string Object everytime. This will affect performance, both memory and time. StringIO's write method will actually modify the string so there will be less overhead by creating new string classes.

app/parse_json_to_object_seq.py
```diff
@@ -1,20 +1,42 @@
...
         self.__controller_method: list[ControllerMethodObject] = []
-        self.__call_nodes: dict[str, dict] = dict()
+        self.__call_nodes: dict[str, CallNode] = dict()
         self.__edges: list = []
-        self.__implicit_parameter_nodes: dict[str, dict] = dict()
+        self.__implicit_parameter_nodes: dict[str, str | int | list[int]] = dict()
+        self.__label_pattern: re.Pattern = re.compile(
+            r"(\[(?P<cond>.*)\] )?(?P<method_name>.*) "
+            r"\((?P<params>.*)?\)( -> (?P<ret_var>.*))?"
+        )
```
DaWanAnOnli: I would personally give a little explanation of the regex, preferably through one or two examples being commented.,ss regex syntax is not the most intuitive thing to read

KingOnyxx: I agree with your feedback regarding this regex syntax and probably should add examples of valid and invalid patterns

MightyZanark: Sure, I'll add some examples later.

PR #24, opened 2025-03-08
Pbi 2 9 write  class functions in views.py

DaWanAnOnli:
- Fix writing MethodObjects and MethodCallObjects by adding fixed parameters and fixed arguments.
- Fixed parameters for ClassMethodObjects: request, instance_name
- Fixed parameter for ControllerMethodObjects: request
- Fixed arguments for ClassMethodCallObjects: request, [instance_name]
- Fixed argument for ControllerMethodCallObjects: request


app/models/diagram.py
```diff
@@ -21,11 +25,27 @@ def to_models_code(self) -> str:
...
+        if self.__parent is not None:
+            res += self.__parent.to_views_code()
+
+        for method in self.__methods:
+            res += method.to_views_code()
+
+        return res
+
     def __str__(self) -> str:
+        """__str__ method for debugging purposes."""
```
MightyZanark, line 48: I did say to add documentation but it should be on our own classes or methods, and not really magic methods, as the only one using these magic methods are us as the developers and users should not need to use it directly. That is my opinion on this anyway, as most magic methods are already documented on what they are supposed to do on the official Python documentation.

MightyZanark (approved):
Overall looks good, keep it up, Joel!

PR #25, opened 2025-03-09
Pbi 2 7 parse misc actions

KingOnyxx:
This pull request includes significant changes to the `app/models`, `app/parse_json_to_object_seq.py`, and `tests` files to improve the functionality and testing of method objects and their interactions. The most important changes include adding equality and representation methods to method objects, enhancing the `print_django_style` method, and updating the parsing logic for JSON data.

Enhancements to JSON parsing:

* [`app/parse_json_to_object_seq.py`](diffhunk://#diff-f6672f8196ff983858af85a8867244eb9da4653b139c58a7c9ff97401b11ec14R2-R42): Added new attributes and methods to `ParseJsonToObjectSeq` class for managing class objects, controller methods, call nodes, edges, implicit parameter nodes, and method calls. Implemented parsing logic for method call sequences with [ ] condition and parse for returning edge.


app/parse_json_to_object_seq.py
```diff
@@ -236,10 +243,26 @@ def parse(self):
...
+                    controller_call.set_method(method)
+                    method.add_call(controller_call)
 
                 elif method not in class_obj.get_methods():
                     class_obj.add_method(method)
 
+                if condition is not None:
+                    method_call = controller_call
+
+                abcde = {
```
KenKomKom, line 255: what is this variable naming, please change it immediately

KingOnyxx: oops! i must've forgotten to change the variable name. Thanks for noticing, i will change this immediately.

KenKomKom:
From what I see, most of the changes made are just adjusting to what linter forces. So, I think it's good to go

PR #26, opened 2025-03-09
PBI 2-5 Create method call object for existing function

NtapSlur:
Finished PBI2-5 Create Method Call Object for Existing Functions
- Implemented for ControllerMethodCallObject and ClassMethodCallObject
-  Create test case for the method call object


DaWanAnOnli:
All good from me.

tests/test_parse_json_to_object_seq.py
```diff
@@ -110,6 +110,327 @@ def test_edge_duplicate_attribute(self):
             parser.parse()
         self.assertEqual(str(context.exception), "Duplicate attribute!")
 
+    def test_positive_class_object(self):
```
DaWanAnOnli, line 113: Why is this test case so long? Is there no way to mock this? Even I have trouble reading this.

tests/test_parse_json_to_object_seq.py
```diff
@@ -110,6 +110,327 @@ def test_edge_duplicate_attribute(self):
...
+            .get_methods()
+            .get_name(),
+            "hasTanggungan",
+        )
+
+        self.assertEqual(
+            len(parsed_value_class_object["ListBuku"].get_methods()[0].get_calls()), 0
+        )
+        self.assertEqual(
+            len(parsed_value_class_object["Buku"].get_methods()[0].get_calls()), 0
```
KenKomKom, line 426: Hey man, why are you doing this? You good? Why is there no mock used in this test file?

tests/test_parse_json_to_object_seq.py
NtapSlur: Since I need to actually compare all of the caller in a method, it's like you need to recursively do it so that I created in each layer of calls and compare all of the methods involved.

tests/test_parse_json_to_object_seq.py
NtapSlur: Because I need to get all of the caller that is being called in the main method so I can assertEqual it accurately. :D

KenKomKom: Ah I see, I guess that is necessary. Okay then

PR #27, opened 2025-03-09
PBI 2: Convert JetUML Sequence Diagram to Code

MightyZanark:
- Added parsing of sequence diagram to our internal object representation
- Added conversion of internal object representation to Django views code


NtapSlur:
I think the code is already good since it already passes the linter check! I also like the Docstring given to the class since it will make it easier to understand the code! Maybe there are some issues in the SonarQube that can be fixed later on, but other than that, the functionality seems fine to me 👍

KingOnyxx:
Looks good to me!

PR #36, opened 2025-03-10
Refactored some missed parts to improve code quality

MightyZanark:
- Changed generic raise Exceptions into ValueError as it is more appropriate for the occasion.
- Changed `assertTrue(a in b)` to `assertIn(a, b)` instead to make it have better error message.
- Merged two if conditions that doesn't really have any benefit being separate.
- Fixed type annotation of parameter `type` in the `set_type` method of `ParameterObject` from `str` to `TypeObject`.

### DIFF
--- a/app/models/diagram.py
+++ b/app/models/diagram.py
@@ -101,12 +101,12 @@ class AbstractRelationshipObject(ABC):
 
     def set_source_class(self, source_class: ClassObject):
         if source_class is None:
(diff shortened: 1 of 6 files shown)


MightyZanark:
Merging as it fixes [6 issues](https://sonarqube.cs.ui.ac.id/dashboard?id=motxt-convert&pullRequest=36)

PR #38, opened 2025-03-12
Handle writing views.py file even when user only uploads class diagram.

DaWanAnOnli:
- Handle writing views.py file even when user only uploads class diagram.
- Generate pass after every class, class method, and controller method definition in models.py and views.py.


KenKomKom:
I think the changes make sens on what the pull request creator is trying to achieve. Okay from me

NtapSlur:
I see that you handled the case when user input is only a class diagram, which is nice and really applicable in real world. I see that you changed some of the test cases, but that's you implemented more so that you need to modify the testcases, which is fine! The rest of the code is already good! Keep up the good work 👍

PR #42, opened 2025-03-14
Setup Bandit and CProfile Profiling

DaWanAnOnli:
- Setup Bandit and CProfile Profiling
- Minor refactoring from Bandit feedback.

I have put the explanation and tutorial on how to use these tools on important-notes channel in Discord

Will be merged at 14/3/2025 21.00 WIB

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ pydantic >=2.10.6,<3.0.0
 starlette >=0.45.3,<0.47.0
 jsonschema >=4.23.0,<4.27.0
 bandit
+snakeviz
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ httpx >=0.28.1,<0.30.0
 pydantic >=2.10.6,<3.0.0
 starlette >=0.45.3,<0.47.0
 jsonschema >=4.23.0,<4.27.0
+bandit
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,3 @@ httpx >=0.28.1,<0.30.0
 pydantic >=2.10.6,<3.0.0
 starlette >=0.45.3,<0.47.0
 jsonschema >=4.23.0,<4.27.0
-bandit
-snakeviz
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -4,3 +4,5 @@ httpx >=0.28.1,<0.30.0
 pydantic >=2.10.6,<3.0.0
 starlette >=0.45.3,<0.47.0
 jsonschema >=4.23.0,<4.27.0
+bandit
+snakeviz
```
MightyZanark, line 8: I suppose these new dependencies are not for the app itself but for us to do profiling? If so then don't put it in this environment, put it in `requirements-dev.txt`. Also pin their versions, either make it absolute using `==` or a range using `>=`, `<` like the others.

KenKomKom:
I assume bandit and CProfiler is used to enhance code cleanliness in our program. Do you perhaps have any technical documentation on how to use it locally or we should discuss this in our group on how to use it so each of us can also use it.

KingOnyxx:
Thanks for implementing this! I think its good to have this profiling tool since our code is parsing data. However, i'm still not sure whats the profiler do when it runs. can it run automatically? or do we have to run it using a command everytime we want to do profiling?

NtapSlur:
It's nice to have Bandit as security issue detector that is Python specific. You also have already explained how to use it in the Discord server so that's good! I also like the idea to use CProfiler to detect methods that have to be refactored so that it can be more efficient and optimized. You also explained how to use CProfiler too which is really great! Thankyou Joel for helping setup these amazing tools!

requirements.txt
DaWanAnOnli: Noted, will do

DaWanAnOnli:
> I assume bandit and CProfiler is used...

Okay I will discuss this the next time I try to use a new library

> Thanks for implementing this! I t...

I have attached the tutorial in the important-notes channle on Discord

requirements.txt
DaWanAnOnli: Done:
https://github.com/placeholder-motxt/convert/pull/42/commits/241d3f6ffdabf1f1c4e6fb7ab659f49bd4abcfa1

PR #43, opened 2025-03-16
Pbi 3 1 modify api

KenKomKom:
In this pull request, I have changed the errors from back end so that it will be pushed towards frontend instead of crashing and only returning 500 internal server. I have also added mroe information regarding the errors so user can have more information on how to fix it


MightyZanark (approved):
I think making the exceptions messages a bit more detailed is a great idea! Approved from me!

DaWanAnOnli (approved):
All good from me

NtapSlur:
It's nice to have improvements for the error message, since this is an issue that we found when doing UAT! It seems good for me! Keep up the good work!

PR #44, opened 2025-03-16
Add Jinja2 version of to_models_code

MightyZanark:
- Minified all the JSON test data to save space.
- Added a Jinja2 templated version of `to_models_code` method on `ClassObject`. This has not been globally implemented yet for the time being as I want some opinions about whether this will be better compared to the regular method.
- Added pep8 naming rules to `ruff.toml` so now the linter will complain when some variable or class name doesn't follow Python's PEP8 rules.


DaWanAnOnli (commented):
There are a lot of changes made in this PR. My question would be is the templated version already working or did you just make the method first? 

Second, how does templating affect performance, whether by space or by time? As how I understand it, templating maybe requires more memory (?), as the template  needs to be loaded all at once first before filling it with actual data.

Third, how much will templating affect the current architecture of the software? As I know it templating is very different algorithmically than conventional string parsing and string generation. If it will dramatically change the architecture, is it something that is worth doing?

That's all from me. I hope to get more insights from your reply. Thanks

MightyZanark:
> There are a lot of changes made in this PR. 

If you take a look, most of this is because of minifying the JSON, removing all whitespace except those that are actually used inside of strings.

> My question would be is the templated version already working or did you just make the method first?

I have not yet integrated it but it more or less works similarly to how it is right now, but with the addition that we can just specify the template name instead of having to chose which method to call.

> Second, how does templating affect performance, whether by space or by time? As how I understand it, templating maybe requires more memory (?), as the template needs to be loaded all at once first before filling it with actual data.

Sure, it definitely will take a bit more memory and a bit more time, and if we just think about *now* it might not look too tempting. But, if we want to expand to other language/framework other than Django that have more boilerplate, using a templating engine might be better as it also improves readability. Frankly, the current `to_models_code` is not that hard to read, but it is a bit hard to see what the resulting file would be. By using templates, people would know how the resulting file structure will look like more easily and if we need to add any changes, we might be able to do it in the template instead of code.

> Third, how much will templating affect the current architecture of the software? As I know it templating is very different algorithmically than conventional string parsing and string generation. If it will dramatically change the architecture, is it something that is worth doing?

It doesn't really change much since at the end of the day the template will produce a Python string, which, is what is already happening. The only real change is the customizability, we can easily support other language/framework by creating a new template and making sure the type mapping is correct for that particular language/framework.

KenKomKom:
I see that there are a lot of changes towards the tests files. Just make sure when merging it with this pull request https://github.com/placeholder-motxt/convert/pull/43 that it merges correctly.

PR #45, opened 2025-03-17
feat: Deployment configs

MightyZanark:
- Removed grafana from compose file as I think it is better to have one for all service compared to one for one service.
- Added prometheus instrumentor to FastAPI app so now prometheus can properly scrape the metrics. I have yet to add custom metrics that could be scraped.
- Added app config so that production env doesn't have API doc and debug=False, while dev and staging have them on.
- Added resource limits to container to maximize resource usage. This limit is arbitrarily chosen for the time being and is open for considerations.


MightyZanark:
> Overall these are good changes, that is to reduce un-needed space and memory in production environment. I do want to ask regarding grafana being one for all services instead of one for every service. Why do think this is a good idea?
> 
> From Adpro I learned that the best practice is for monitoring to be done at each service, so that the metrics being monitored are specific, and this also complies with the Single Responsibility Principle, that is, one monitoring endpoint for each service.

That is quite different from what I have been told from my lecturer at AdvPro. My lecturer said that the Grafana itself should just be kept at one but it has multiple dashboards containing the metrics of each services, and I agree with him in that we don't have that many services to really make the best use of 1 Grafana container for each service. If needed, we can have replicas of the Grafana container to make sure own instance being down makes us unable to monitor our services, but having a Grafana dashboard for each and every service still feels a bit excessive to me. I was also considering making one centralized Prometheus instance, but refrained from doing so as I think it is still fine to have one Prometheus instance for each service.

> Sure, having only one monitoring endpoint may save memory and computational power, leading to better performance. However, architecturally considering we have many repositories, does that mean the monitoring will only be done for FE (which is connected to all services) or will we still monitor each service in detail? If the former is used, I believe we are missing some key metrics to be monitor. If the latter is used, how expensive is it for one monitoring endpoint to inspect services provided by other repositories?

We just need to set up the one Grafana instance to take metrics from all the Prometheus instances across services, so we can still monitor every service and not just one.

DaWanAnOnli:
Ah I see that make sense.. I think I was mixed up between setting up Grafana and Prometheus. So as I understand it, Prometheus will still be set up for all services, but there will be one centralized Grafana endpoint to pick up the metrics from the Prometheuses. That makes sense. Otherwise I was a bit confused as to how you would get the metrics from all services by setting up grafana and prometheus from one service only.

Cheers.

KenKomKom:
Looks good to me. Also based on what I've read from the discussion, it aligns with what I also understand from advance programming. So, I'm good to go

PR #46, opened 2025-03-17
Bug-Fix: 1 - Implement More Flexible Method That Is Allowed

NtapSlur:
Changed the rule for method name so that it can be more flexible:
- Before, it only allows method with this pattern {method name} ({parameters}), but now it supports methods with this pattern {method name}({parameters}) or {method name} ({parameters})


KenKomKom:
Looks good to me, I approve merge

tests/test_parse_json_to_object_seq.py
```diff
@@ -449,23 +449,22 @@ def test_positive_method_call_object(self):
...
-            "Wrong label format 'doA()' on sequence diagram \n"
-            "please consult the user manual document on how to name parameters",
+            parser.get_class_objects()["ABC"].get_methods()[0].get_name(), "doA"
         )
 
+    def test_invalid_edge_label_format(self):
+        # Invalid label format should throw exceptions, examples:
+        # `doA()` -> have to be `doA ()`
+        # `[] doB ()->bval` -> have to be `[] doB () -> bval`
+
```
DaWanAnOnli, line 467: Isn't the docstring supposed to be changed as now doA() is valid?

DaWanAnOnli (changes requested):
There is just one detail putting me off that the docstring for test_invalid_edge_label_format is still unchanged. You should change it for consistency as this may become maintenance issue in the future

KingOnyxx:
Thank you for addressing the flexibility of method naming. The changes look good overall. I have reviewed the modifications and tested the new method patterns. Other than the comment @DaWanAnOnli's mentioned, i think it looks good to me.
Great job!

tests/test_parse_json_to_object_seq.py
NtapSlur: Oops it seems that i missed that part, gonna fix it!

PR #47, opened 2025-03-19
Pbi 4 3 create requirements.txt file

KenKomKom:
In this pull request, I have created a FileElement child class that writes a set of Requirements.txt for user to install before running django


app/models/elements.py
```diff
@@ -132,3 +134,46 @@ def add_class_method(self, class_method_object: ClassMethodObject):
...
+        result = StringIO()
+        requirements = [
+            "django==4.1.4",
+            "gunicorn==20.1.0",
+            "whitenoise==6.9.0",
+            "psycopg2==2.9.5",
+            "pytest==7.2.2",
+            "pytest-django==4.5.2",
+            "pytest-cov==4.0.0",
+        ]
```
MightyZanark, line 174: I don't think version pinning here is the best idea since that means our resulting project will always be "old". I think its better to just leave the versions for the user to pin it themselves later on what they think is best. Pinning it now might make the user have to search old documentation of those dependencies instead of just going to the latest one.

Also the `d` in `django` needs to be [uppercase](https://pypi.org/project/Django/).

MightyZanark (changes requested):
Overall looks good, just some opinion on the requirements.

app/models/elements.py
```diff
@@ -132,3 +134,46 @@ def add_class_method(self, class_method_object: ClassMethodObject):
...
+
+    async def write_to_file(self, path: str) -> None:
+        file = path + "/requirements.txt"
+
+        to_be_print = self.print_django_style()
+
+        async with await anyio.open_file(file, "w") as f:
+            await f.write(to_be_print)
+        print("done writing", file)
+        return file
```
DaWanAnOnli, line 159: Currently you have this async method inside the RequirementsElements class. However, this is inconsistent with other classes, as ModelsElements and ViewsElements do not have this method. Currently the async method is called directly from main.py. There are two options here, both having to do with software architecture
1. Move this method to main.py file, to make it consistent with other subclasses of FileElements
2. Move the async functions in main.py to the FileElements class. In this case main.py will only be calling methods, while the method definition is contained in FileElements class

Either way, we need consistency in the structure of these classes. Either all FileElements have write_to_file, or not at all

DaWanAnOnli (changes requested):
A small suggestion that affects project structure.

app/models/elements.py
KenKomKom: So I should just leave the version as blank?

PR #50, opened 2025-03-19
Pbi 6 2 set up urls.py

KenKomKom:
In this pull request I have set up an element class to write urls.py based on the input of a list of files. It will generate a file named urls.py based on the path given. 
You can see how to use it on main.py where there is a commented code showing how to generate a urls.py on folder /app.

Cheers!


app/models/elements.py
```diff
@@ -132,3 +135,19 @@ def add_class_method(self, class_method_object: ClassMethodObject):
...
+class UrlsElement(FileElements):
+    async def write_file(self, classes: list, path: str) -> str:
+        file = path + "/urls.py"
+
+        to_be_print = self.print_django_style(classes)
+
+        async with await anyio.open_file(file, "w") as f:
+            await f.write(to_be_print)
+
+        return file
```
DaWanAnOnli, line 148: Similar comment to your previous PR. Currently you have this async method inside the UrlsElements class. However, this is inconsistent with other classes, as ModelsElements and ViewsElements do not have this method. Currently the async method is called directly from main.py. There are two options here, both having to do with software architecture

Move this method to main.py file, to make it consistent with other subclasses of FileElements
Move the async functions in main.py to the FileElements class. In this case main.py will only be calling methods, while the method definition is contained in FileElements class
Either way, we need consistency in the structure of these classes. Either all FileElements have write_to_file, or not at all.

DaWanAnOnli (changes requested):
Minor suggestion regarding project structure

app/models/elements.py
```diff
@@ -132,3 +135,19 @@ def add_class_method(self, class_method_object: ClassMethodObject):
...
+
+    def print_django_style(self, classes: list = []) -> str:
+        template = Template(open("./app/templates/urls.py.j2").read())
```
MightyZanark, line 151: Please use the `render_template` function that is already defined in `utils.py`. You can pass the context as `{"classes": classes}` since I forgot to add kwargs as the parameter of that function.

MightyZanark (changes requested):
Other than the comments from Joel and the one I have, everything else looks good!

PR #51, opened 2025-03-20
Pbi 5 1 generate create page

DaWanAnOnli:
- Generated Jinja html page for create page. For now this will only return string. Each generated class will have its own page.
- Generated forms.py file using Jinja. Similar to the html files, this will also return string. There will only be one forms.py file for all classes.

### DIFF
--- a/app/generate_frontend/create/generate_create_page_django.py
+++ b/app/generate_frontend/create/generate_create_page_django.py
@@ -1,9 +1,7 @@
-from jinja2 import Environment, PackageLoader
-
-from app.models.diagram import ClassObject, FieldObject
+from app.models.diagram import ClassObject
 from app.models.elements import ModelsElements
+from app.utils import render_template
 
-env = Environment(loader=PackageLoader("app"))
 
 def generate_html_create_pages_django(models_elements: ModelsElements) -> list[str]:
     result = []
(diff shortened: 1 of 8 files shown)


app/generate_frontend/create/generate_create_page_django.py
```diff
@@ -0,0 +1,47 @@
...
+    field2.set_name("age")
+    field3 = FieldObject()
+    field3.set_name("is_alive")
+    class_object.add_field(field1)
+    class_object.add_field(field2)
+    class_object.add_field(field3)
+    models_element = ModelsElements("Person")
+    models_element.add_class(class_object)
+    models_element.add_class(class_object)
+    print(generate_forms_create_page_django(models_element))
```
KingOnyxx, line 47: is this block of code necessary?
also i think you forgot to remove the print statement

KingOnyxx (changes requested):
Thank you for implementing this. Before merging, i would consider removing the block of code i commented. Other than that, i approve to merge

app/generate_frontend/create/generate_create_page_django.py
```diff
@@ -0,0 +1,47 @@
+from jinja2 import Environment, PackageLoader
+
+from app.models.diagram import ClassObject, FieldObject
+from app.models.elements import ModelsElements
+
+env = Environment(loader=PackageLoader("app"))
```
MightyZanark, line 6: Please use the one already defined in `utils.py` instead of creating another instance of the Environment. In `utils.py` there is also already a `render_template` method that does grabs the template and renders it with the given context.

MightyZanark (changes requested):
Other than the one comment I give, it looks good to go!

app/generate_frontend/create/generate_create_page_django.py
DaWanAnOnli: Thankyou for the suggestion. I will make these changes

app/generate_frontend/create/generate_create_page_django.py
DaWanAnOnli: Nice catch. Yes, this is my oversight. I will remove these immediately

DaWanAnOnli:
I have made the changes suggested by @MightyZanark  and @KingOnyxx . Thankyou

PR #52, opened 2025-03-20
Pbi 5 5 views.py for create page

DaWanAnOnli:
- Generated Jinja views.py file for create page. For now this will only return string. There will only be one views.py file in total.

Deadline: Friday, 21 March 2025, 08.00 WIB

### DIFF
--- a/app/templates/create_page_views.py.j2
+++ b/app/templates/create_page_views.py.j2
@@ -1,0 +1,16 @@
+from django.shortcuts import render, redirect
+from .forms import *
+from .models import *
+
+{% for class in classes %}
+def create_{{ class.name_snake }}(request):
+    if request.method == 'POST':
+        form = {{ class.name }}Form(request.POST)
+        if form.is_valid():
+            form.save()
+            return redirect('view_{{ class.name_snake }}')
+    else:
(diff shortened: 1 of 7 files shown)


app/templates/create_page_views.py.j2
```diff
@@ -0,0 +1,16 @@
+from django.shortcuts import render, redirect
+from .forms import *
+from .models import *
+
+{% for class in classes %}
+def create_{{ class.name_snake }}(request):
```
KenKomKom, line 6: Ah i see that you've implemented snake case generation for the classes of the user. the urls.py generation will need to be updated then. How do you think we should do it as urls.py is also on pull request as is this one

KingOnyxx (approved):
Your implementations looks very well made, i approve to merge!

app/templates/create_page_views.py.j2
DaWanAnOnli: I will PR this to dev and then you can pull it from there.

DaWanAnOnli:
Generated Jinja views.py file for create page. For now this will only return string. There will only be one views.py file in total.

PR #57, opened 2025-03-21
BUG-FIX-2-Fix Method Call Order

NtapSlur:
Parser for sequence diagram now process Method Call Object sequentially based on the order of occurence, not based on sorted id.


KenKomKom:
This is quite small of a change, and I think it looks okay. I approve this PR to be merged

KingOnyxx (approved):
After reviewing this PR, i believe we can approve this to merge to dev

MightyZanark (approved):
Seems like everything still working as intended after this small change. I approve of this change.

PR #58, opened 2025-03-23
Pbi 5 7 edit class diagram parsing to support public private character to specify public classes

KenKomKom:
In this pull request, I've managed to add the following features:
public private class parsing
is_public attribute writing to django file

### DIFF
--- a/app/models/diagram.py
+++ b/app/models/diagram.py
@@ -107,7 +107,6 @@ class ClassObject:
         res = StringIO()
         for attribute in self.__fields:
             res.write("\t" + attribute.to_models_code() + "\n")
-        res.write(f"\tis_public={self.__is_public}\n")
         return res.getvalue()
 
     def __get_relationships_to_code(self) -> str:
--- a/app/models/diagram.py
+++ b/app/models/diagram.py
@@ -107,6 +107,7 @@ class ClassObject:
         res = StringIO()
         for attribute in self.__fields:
             res.write("\t" + attribute.to_models_code() + "\n")
+        res.write(f"\tis_public={self.__is_public}\n")
         return res.getvalue()
 
     def __get_relationships_to_code(self) -> str:
--- a/app/models/diagram.py
+++ b/app/models/diagram.py
@@ -19,10 +19,11 @@ class ClassObject:
         self.__fields: list[FieldObject] = []
         self.__methods: list[ClassMethodObject] = []
         self.__relationships: list[AbstractRelationshipObject] = []
+        self.__is_public: bool = False
 
         self.__id: int
 
-    def get_fields(self) -> list[FieldObject]: # pragma: no cover
+    def get_fields(self) -> list[FieldObject]:  # pragma: no cover
         return self.__fields
 
     def to_models_code(self) -> str:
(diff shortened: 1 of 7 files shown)


app/parse_json_to_object_class.py
```diff
@@ -45,6 +45,16 @@ def parse_classes(self) -> list:
 
             object_name = object["name"]
 
+            print("zczczc", object_name, object_name[0])
```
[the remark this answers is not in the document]
MightyZanark: +1 to this

app/parse_json_to_object_class.py
```diff
@@ -45,6 +45,16 @@ def parse_classes(self) -> list:
...
+            if bool(re.match(r"([\+\-]) ?(?P<class_name>\w*)", object_name)):
+                if object_name[0] == "+":
+                    class_obj.set_is_public(True)
+                else:
+                    class_obj.set_is_public(False)
+                object_name = object_name[1:].strip()
```
MightyZanark, line 54: You might as well use the match object that is given by `re.match` to extract the public/private and the class name. Also, please compile the pattern and put it as global constant or class constant since this pattern will be used every time parsing occurs.

MightyZanark (changes requested):

MightyZanark:
Overall looks good, just might want to use the match object (if it matches) instead of turning it into a boolean and then parsing manually.

app/models/diagram.py
```diff
@@ -103,6 +107,7 @@ def __get_attributes_to_code(self) -> str:
         res = StringIO()
         for attribute in self.__fields:
             res.write("\t" + attribute.to_models_code() + "\n")
+        res.write(f"\tis_public={self.__is_public}\n")
```
DaWanAnOnli, line 110: Does this need to be written directly into models.py? I thought is_public is only needed to determine whether the FE needs to be made. I think it should not affect the models.py code at all.

DaWanAnOnli (changes requested):
A minor suggestion on whether is_public needs to be written in generated models.py file

app/parse_json_to_object_class.py
```diff
@@ -45,6 +45,16 @@ def parse_classes(self) -> list:
 
             object_name = object["name"]
 
+            print("zczczc", object_name, object_name[0])
+            if bool(re.match(r"([\+\-]) ?(?P<class_name>\w*)", object_name)):
```
KingOnyxx, line 49: Can you give docstring for this so future developer can understand what kind of string can be accepted by this regex?

KingOnyxx (changes requested):
Having reviewed at the changes you've made, i think its better to merge into PBI 5 first before merging into dev. 
Also, you should add changes from me and the other commenters' feedbacks.

app/models/diagram.py
KenKomKom: After thinking about it again, yeah I don't think it will be necessary to add the is_public to models.

app/parse_json_to_object_class.py
KenKomKom: okay, will do

PR #60, opened 2025-03-24
PBI:4-4-Create Script to Run the Django Project

NtapSlur:
Added the jinja template for the MoTxT to generate script, both .sh and .bat to run the Django Project. The script will do these things:
1. Check whether env folder exists
a. If yes, then turn on the env
b. If no, then create a new virtual env and turn on the env

2. Install the requirements.txt

3. Do migration for the database

4. Run the server


app/models/elements.py
```diff
@@ -171,3 +173,29 @@ def print_django_style(self) -> str:
...
+        of .sh (Linux & MacOS)
+        """
+        super().__init__(file_name)
+
+    def print_django_style(self) -> str:
+        return render_template("scripts/run.sh.j2", {})
+
+
+class RunBatScriptElements(FileElements):
+    def __init__(self, file_name: str):
```
KenKomKom, line 192: This function is not necessary, if no constructor is created, python will implicitly run parent's constructor. So we can just remove this __init__ method

app/templates/scripts/run.bat.j2
```diff
@@ -0,0 +1,30 @@
+@echo off
+
+REM Check if env folder exists
+if exist env (
```
KenKomKom, line 4: so I assume if the user has a env folder not named env it will create another one? that seems redundant, I think we should at least check the common env folder names like venv, and so on

KenKomKom:
Besides the small suggestions I made, I think overall it is still okay to be merged as is

app/templates/scripts/run.bat.j2
NtapSlur: I think that since the intention is to minimalize user's task on starting the Django app and since the Django project is empty with no virtual env, it seems fine to let the virtual env folder as 'env' only. For the case like if user creates his/her own virtual env, i think i can add parameter to the script for the path of the env itself so that it creates more flexibility for the custom env. Thankyou for the suggestion and feedback by the way!

app/models/elements.py
NtapSlur: Okay ken. Thankyou for the feedback!

KingOnyxx (approved):
If you dont use any context for Jinja, i think its better to just write the file by itself (not using jinja for templating). Please give me reasons why you chose to use jinja.
Other than that, I believe we can proceed with merging this pull request.

NtapSlur:
> If you dont use any context for Jinja, i think its better to just write the file by itself (not using jinja for templating). Please give me reasons why you chose to use jinja. Other than that, I believe we can proceed with merging this pull request.

Since we are using Jinja as templating, I think that it is best to keep using Jinja for like standarization of file templating so that the behavior of how file is generated is still the same. In performance side, I think there are no difference too in using Jinja and write the file itself.

app/templates/scripts/run.bat.j2
NtapSlur: Implemented this!

MightyZanark:
> > If you dont use any context for Jinja, i think its better to just write the file by itself (not using jinja for templating). Please give me reasons why you chose to use jinja. Other than that, I believe we can proceed with merging this pull request.
> 
> Since we are using Jinja as templating, I think that it is best to keep using Jinja for like standarization of file templating so that the behavior of how file is generated is still the same. In performance side, I think there are no difference too in using Jinja and write the file itself.

Performance wise, you have to load the file as a Jinja2 Template, which definitely takes more memory than if it is just a regular static file. I don't think the use of `render_template` is appropriate here as it will just result in more function calls without any real benefit. Templating by itself is supposed to be done for files that can have different contents depending on the context. If a file is really just a static file, there is no reason to use a templating engine to process said file.

PR #61, opened 2025-03-24
Pbi 4 2 create django app generated django project

KingOnyxx:
This pull request introduces several changes to the `app/main.py` file, adds new templates for Django apps, and includes new tests to validate the functionality. The most important changes include the addition of the `create_django_app` function, updates to Django app templates, and new test cases for generating Django apps.

### New Functionality:
* Added `create_django_app` function to generate a Django app within a project zip file. This function validates the project and app names, reads template files, and writes them to the zip file.

### Template Updates:
* Added new Django app template files for `admin.py`, `apps.py`, `models.py`, `tests.py`, and `views.py`. These files contain basic structure and placeholders for further development. [[1]](diffhunk://#diff-f6f1e083e0cc1c8dfcfd086a47ceb42b659c982a1f4592dfb0dac8d85cd932e0R1) [[2]](diffhunk://#diff-e5794d55a6aa5de209161b69905cb706bd77abdb71cfdb23e128453b6c06ada8R1-R6) [[3]](diffhunk://#diff-ee768f7ea6e9c1ae1598d072a440bac4458c62017a4cebcf79c933a1ed9a114dR1) [[4]](diffhunk://#diff-0c0290d98c6dcbff908fc107bd36f8fc5de8d127d29944c3296da346a78008eeR1) [[5]](diffhunk://#diff-09ee35fc942a3e58d32ca690b6d23c1241271878ad1ffe15d11a568c33e56a59R1)

### Testing Enhancements:
* Updated `tests/test_render_project_django.py` to import the new `create_django_app` function.
* Added new test cases in `tests/test_render_project_django.py` to validate the positive and negative scenarios for generating Django apps, including checks for invalid names and non-existent zip files.


DaWanAnOnli:
Are the .py files in the templates/django_app directory supposed to be Jinja template files? If so, I would recommend naming them this way: <file_name>.py.j2 , simply to differentiate them from the actual MoTxT source code.

KingOnyxx:
> Are the .py files in the templates/django_app directory supposed to be Jinja template files? If so, I would recommend naming them this way: <file_name>.py.j2 , simply to differentiate them from the actual MoTxT source code.

at first it was not suppoed to be a jinja file, because its just a bare django app that doesnt need templating.
but after some consideration, i think im going to refactor them to use jinja instead.

app/main.py
```diff
@@ -207,3 +212,27 @@ def create_django_project(project_name: str) -> list[str]:
...
     zipf.close()
     return files
+
+
+def create_django_app(project_name: str, app_name: str) -> list[str]:
+    file_names = []
+    if not is_valid_python_identifier(app_name):
+        raise ValueError("App name must not contain whitespace or number!")
+    if not is_valid_python_identifier(project_name):
+        raise ValueError("Project name must not contain whitespace or number!")
```
MightyZanark, line 222: The `is_valid_python_identifier` allows numbers to be in the string, so the error message is not quite correct. The one that is not allowed is a dash `-`.

app/main.py
```diff
@@ -207,3 +212,27 @@ def create_django_project(project_name: str) -> list[str]:
...
+            content = f.read()
+        # [TODO] place to add stuff in the file
+
+        # check if project zip file exist
+        if not os.path.exists(f"{project_name}.zip"):
+            raise FileNotFoundError(f"File {project_name}.zip does not exist")
+        with zipfile.ZipFile(f"{project_name}.zip", "a") as zipf:
+            if file == "__init__.py":
+                zipf.writestr(f"{app_name}/migrations/__init__.py", content)
+            zipf.writestr(f"{app_name}/{file}", content)
```
MightyZanark, line 236: Why not just make the context the top most block instead of the for loop? This way you don't need to reopen the context every time to write file.

tests/test_render_project_django.py
```diff
@@ -124,3 +124,67 @@ def test_render_project_django_template_negative_with_existing_folder(self):
...
+        zipfile_path = "test_main.zip"
+        open_zip = zipfile.ZipFile(zipfile_path, "r")
+        files = open_zip.namelist()
+
+        for file in files:
+            filename = file.split("/")[-1]
+            if filename in os.listdir("app/templates/django_app"):
+                if "migrations" in file:
+                    continue
+                print(open_zip.extract(file, os.path.join("tests", "testdata")))
```
MightyZanark, line 151: Don't forget to remove print statements.

MightyZanark (changes requested):
Overall it looks good, but there are some parts where I would like you to change as denoted by the other comments.

MightyZanark:
> > Are the .py files in the templates/django_app directory supposed to be Jinja template files? If so, I would recommend naming them this way: <file_name>.py.j2 , simply to differentiate them from the actual MoTxT source code.
> 
> at first it was not suppoed to be a jinja file, because its just a bare django app that doesnt need templating. but after some consideration, i think im going to refactor them to use jinja instead.

If they are static files, which doesn't really get any benefit of using through Jinja2, I think we can use some other extension instead of `.j2`. Maybe we should discuss it later.

PR #62, opened 2025-03-24
Pbi 5 2 generate html read page

DaWanAnOnli:
- Generate html file to display all public models.

### DIFF
--- a/app/generate_frontend/read/generate_read_page_django.py
+++ b/app/generate_frontend/read/generate_read_page_django.py
@@ -5,7 +5,10 @@ from app.utils import camel_to_snake, render_template
 
 def generate_html_read_pages_django(models_elements: ModelsElements) -> list[str]:
     if not isinstance(models_elements, ModelsElements):
-        raise (TypeError("argument must be of type ModelsElements!"))
+        raise TypeError(
+            f"Expected type ModelsElements, got {type(models_elements)} instead"
+        )
+    # Consider changing the code below to avoid memory explosion
     result = []
     for class_object in models_elements.get_classes():
         result.append(generate_html_read_page_django(class_object))
@@ -14,13 +17,11 @@ def generate_html_read_pages_django(models_elements: ModelsElements) -> list[str
 
 def generate_html_read_page_django(class_object: ClassObject) -> str:
     if not isinstance(class_object, ClassObject):
-        raise (TypeError("argument must be of type ClassObject!"))
+        raise TypeError(f"Expected type CLassObject, got {type(class_object)} instead")
     context = {
         "class_name": class_object.get_name(),
         "class_snake": camel_to_snake(class_object.get_name()),
     }
-    fields = []
-    for field in class_object.get_fields():
-        fields.append({"name": field.get_name()})
+    fields = [{"name": f.get_name()} for f in class_object.get_fields()]
     context["fields"] = fields
     return render_template("read_page_django.html.j2", context)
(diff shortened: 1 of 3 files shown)


app/templates/read_page_django.html.j2
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>{{ class_name }} List</title>
+</head>
```
MightyZanark, line 7: Since I think we will be using this format for most of our HTML templates, I think it would be better to split the template into smaller templates that extends another and have blocks, similar to how we did it in Django.

app/generate_frontend/read/generate_read_page_django.py
```diff
@@ -0,0 +1,26 @@
+from app.models.diagram import ClassObject
+from app.models.elements import ModelsElements
+from app.utils import camel_to_snake, render_template
+
+
+def generate_html_read_pages_django(models_elements: ModelsElements) -> list[str]:
+    if not isinstance(models_elements, ModelsElements):
+        raise (TypeError("argument must be of type ModelsElements!"))
```
MightyZanark, line 8: Why the extra brackets wrapping TypeError? A typo?

app/generate_frontend/read/generate_read_page_django.py
```diff
@@ -0,0 +1,26 @@
...
+    result = []
+    for class_object in models_elements.get_classes():
+        result.append(generate_html_read_page_django(class_object))
+    return result
```
MightyZanark, line 12: This can potentially explode memory usage but I don't have much idea on how to work around it right now, maybe you want to mark it with something so that we can remember to find a better solution for it.

MightyZanark (changes requested):
There are some things I would like to be changed before merging, but overall I think its good.

Not0nline:
halo joel, untuk kodenya sebagian besar sudah bagus, mungkin berikut beberapa hal kecil yang kamu bisa ganti

1.  **Hapus Redundant Parentheses**  
- Sebelum
   raise (TypeError("..."))
   
- Sesudah
   raise TypeError("...")

2. List Comprehension untuk Fields
- Sebelum
fields = []
for field in class_object.get_fields():
    fields.append({"name": field.get_name()})

- Sesudah
fields = [{"name": f.get_name()} for f in class_object.get_fields()]

3. Pesan Error Lebih Spesifik
-  Sebelum
raise TypeError("argument must be of type ModelsElements!")

- Sesudah
raise TypeError(
    f"Expected {ModelsElements.__name__}, got {type(models_elements).__name__}"
)

4. Colspan Dinamis di Template
- Sebelum 
<td colspan="3">...</td>

- Sesudah
<td colspan="{{ fields|length }}">...</td>

5. Perbaikan URL Tag Django
- Sebelum
{% url 'create_{{ class_snake }}' %}

- Sesudah
{% url 'create_'|add:class_snake %}

tapi untuk kode yang kamu kasih sudah oke tanpa issue2 major sehingga sudah bisa dimerge, thank you

app/templates/read_page_django.html.j2
DaWanAnOnli: Okay, however I think this will better be done directly in the PBI-5 branch, when all the templates are already put in one place. However I do think this is something that absolutely needs to be done. Thanks for the feedback.

app/generate_frontend/read/generate_read_page_django.py
DaWanAnOnli: Yes. Thanks for the catch. I will fix this immediately

app/generate_frontend/read/generate_read_page_django.py
DaWanAnOnli: Noted. I will give comments regarding this issue

PR #64, opened 2025-03-24
Pbi 5 12 create views.py function for delete

KenKomKom:
On this PR, I've managed to do:

- Make a template for writing functions to delete object on django
- Make a backend for a POST request to handle object deletion on django

### DIFF
--- a/app/templates/delete_page_views.py.j2
+++ b/app/templates/delete_page_views.py.j2
@@ -1,0 +1,14 @@
+{% for class in classes %}
+def delete_{{ class.name_snake }}(request):
+    if request.method == 'POST':
+        data = json.loads(request.body)
+        object_id = data.get("id")
+
+        if object_id is None:
+            return JsonResponse({"error": "ID is required"}, status=400)
+
+        obj = {{class.name}}.objects.get(pk=object_id)
+        obj.delete()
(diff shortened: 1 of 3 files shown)


app/templates/delete_page_views.py.j2
```diff
@@ -0,0 +1,14 @@
...
+    if request.method == 'POST':
+        data = json.loads(request.body)
+        object_id = data.get("id")
+
+        if object_id is None:
+            return JsonResponse({"error": "ID is required"}, status=400)
+
+        obj = {{class.name}}.objects.get(pk=object_id)
+        obj.delete()
+        return JsonResponse({"message": f"Object with ID {object_id} deleted successfully."})
```
DaWanAnOnli, line 12: Now this is interesting. Does this mean that we must add an ID field for every class in MoTxT? As I don't think we currently have it right now. (Or am i wrong?)
This will actually affect a lot of the previous code and code architecture, as this means users cannot add and ID field in their class diagrams, as the ID field will be automatically generated. I believe this is something we need to discuss further.

DaWanAnOnli (commented):
A discussion regarding the ID field of models

app/templates/delete_page_views.py.j2
KenKomKom: Django automatically gives id to class if the class has no primary keys

PR #66, opened 2025-03-25
Pbi 5 6 create views.py function for read

DaWanAnOnli:
- Completed PBI-5: 6
- Added base views as base template for views.py file.


app/templates/read_page_views.py.j2
```diff
@@ -0,0 +1,7 @@
+{% for class in classes %}
+def {{ class.class_snake }}_list(request):
```
KenKomKom, line 2: can you change this to get_class.snake_name. Because in the urls.py, I've implemented it to write get_class.snake_name as the read function from views.

DaWanAnOnli: okay.. noted

PR #67, opened 2025-03-25
PBI-4-5-Replace Default models.py With The Parsed One

NtapSlur:
Added method to get the django style from the parsed models.py and change the default models.py into the parsed models.py
- Create new method render_model
- Modify method create_django_app to support the parsed models.py

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -215,16 +215,21 @@ def create_django_project(project_name: str) -> list[str]:
     return files
 
 
-def create_django_app(
-    project_name: str, app_name: str, models: str = None
-) -> list[str]:
-    file_names = []
+def validate_django_app(project_name: str, app_name: str):
     if not is_valid_python_identifier(app_name):
         raise ValueError("App name must not contain whitespace!")
     if not is_valid_python_identifier(project_name):
         raise ValueError("Project name must not contain whitespace!")
     if not os.path.exists(f"{project_name}.zip"):
         raise FileNotFoundError(f"File {project_name}.zip does not exist")
+
+
+def create_django_app(
+    project_name: str, app_name: str, models: str = None
+) -> list[str]:
+    file_names = []
+
+    validate_django_app(project_name, app_name)
 
     with zipfile.ZipFile(f"{project_name}.zip", "a") as zipf:
         for file in os.listdir("app/templates/django_app"):
--- a/app/main.py
+++ b/app/main.py
@@ -215,7 +215,9 @@ def create_django_project(project_name: str) -> list[str]:
     return files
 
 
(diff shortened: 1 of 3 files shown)


app/main.py
```diff
@@ -236,7 +243,22 @@ def create_django_app(project_name: str, app_name: str) -> list[str]:
                 zipf.writestr(f"{app_name}/{file_name}", template)
                 file_names.append("apps.py")
             else:  # file that use txt file
-                if file == "__init__.txt":
+                """
+                Check if the current file is models.txt and if yes, render based on
+                the parsed value in the parameter 'models'
+                """
+                if file == "models.txt" and models is not None:
+                    if models != "":
```
DaWanAnOnli, line 251: I'm a bit confused here. I don't recall us ever using or processing any file named models.txt. Did you mean models.py, models_file.class.jet or models.py.j2?

DaWanAnOnli (commented):
A comment regarding file processing. You use a file named models.txt, which is unfamiliar to me. I don't think we've ever mentioned this particular file name in our project.

app/main.py
NtapSlur: The file really is models.txt. It is used in the create_django_app to create empty Django app. You can find the file in the templates/django_app/models.txt. It was made by Clarence to set empty models.py

app/main.py
```diff
@@ -236,7 +243,22 @@ def create_django_app(project_name: str, app_name: str) -> list[str]:
...
+                        zipf.writestr(f"{app_name}/models.py", models)
+                    else:
+                        with open(
+                            os.path.join("app", "templates", "django_app", file), "r"
+                        ) as f:
```
MightyZanark, line 256: You can just declare the current directory as a constant in this file like `CUR_DIR = os.path.dirname(os.path.realpath(__file__))` and then declare another one for template directory or the django app file directory like `TEMPLATE_DIR = os.path.join(CUR_DIR, "templates")` or `DJANGO_APP_DIR = os.path.join(CUR_DIR, "templates", "django_app")` so that you don't join a lot of paths at once multiple times, but only about 2 path multiple times.

MightyZanark (approved):
Looks good. Just a small request to make the directory path a constant instead of having to build it every time we need to write a file inside it.

KingOnyxx (approved):
After reviewing this PR, i have a few suggestion for you to do:

- Ensure that the `validate_django_app` function correctly validates the project and app names and the existence of the project zip file.

- Verify that the new `render_model` function accurately converts JSON class diagram content into Django models.

Overall, these changes enhance the flexibility and robustness of the create_django_app function. Great work on the refactor and the addition of new capabilities!

PR #68, opened 2025-03-26
feat: Edit page views for frontend generation

MightyZanark:
- Added function to generate edit page views that will generate edit functions for any class that have fields. This has not yet taken into account inheritance.
- Refactored `render_template` function parameters so that is it more in line with the `render` method of `Template` in Jinja. This is just a parity change and I have confirmed everything still works as intended.


DaWanAnOnli (approved):
All good from me

KingOnyxx (approved):
After reviewing, i believe i can approve of this pull request

PR #71, opened 2025-03-26
PBI-4-6-Replace Views.py With The Generated One

NtapSlur:
Added logic to replace the default views.py with the generated one
- fetch_data() is implemented to process all the .jet file first, and then render_model and render_views will get the value of the models.py and views.py
- Added new method to get return variable name since somehow the original set_return_var_name is broken :/
- Modified previous tests that is only for models so that supports tests for views too!


DaWanAnOnli (changes requested):

app/models/methods.py
```diff
@@ -290,6 +294,7 @@ def print_django_style(self) -> str:
 
         """
         result = StringIO()
+        print(self.get_return_var_name() + " HEHE " + self.__method.get_name())
```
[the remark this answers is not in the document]
NtapSlur: Oops, seems that forgot to delete the print for debugging stuff. Gonna fix it soon!

app/models/methods.py
```diff
@@ -246,6 +246,10 @@ def get_arguments(self) -> list[ArgumentObject]:
     def add_argument(self, argument: ArgumentObject):
         self.__arguments.append(argument)
 
+    # Method yang dibuat akibat kejadian ghaib
```
KingOnyxx, line 249: There are code comments in non-English language which might be unclear to collaborators not familiar with the language. Consistency in language across the codebase is important for maintainability.

app/main.py
```diff
@@ -248,15 +248,13 @@ def create_django_app(
...
                 if file == "models.txt" and models is not None:
-                    if models != "":
-                        zipf.writestr(f"{app_name}/models.py", models)
-                    else:
-                        with open(
-                            os.path.join("app", "templates", "django_app", file), "r"
-                        ) as f:
-                            content = f.read()
-                            file_name = file.replace(".txt", ".py")
-                            zipf.writestr(f"{app_name}/{file_name}", content)
```
KingOnyxx, line 259: why delete this block of code? this is supposed to be for the other files in django_app folder

KingOnyxx (changes requested):
After reviewing this PR changes, there are a few comments i left for you to change before merging this PR.

app/main.py
NtapSlur: The deleted block of code is only if the file is models.txt as you can see in the conditional. I deleted the previous one for the models because the parser wont return empty string. The rest of the files will be generated in the else condition

KingOnyxx: Oh i see, i thought that you deleted my code. That's my bad!

app/models/methods.py
NtapSlur: Thankyou for the suggestion! Gonna fix it

PR #72, opened 2025-03-27
Pbi 5 9 generate landing page

DaWanAnOnli:
Finished PBI-5-9:
Added landing page generation. Landing page will contain the names of all public classes. The names themselves are links, so if you click them you will be directed to the corresponding class' read page.

### DIFF
--- a/app/generate_frontend/generate_landing_page.py
+++ b/app/generate_frontend/generate_landing_page.py
@@ -1,0 +1,9 @@
+from app.utils import render_template
+
+
+def generate_landing_page_html() -> str:
+    return render_template("landing_page.html.j2", {})
+
+
+def generate_landing_page_views() -> str:
+    return render_template("landing_page_views.py.j2", {})
--- a/app/templates/base_views.py.j2
+++ b/app/templates/base_views.py.j2
@@ -1,3 +1,4 @@
 from django.shortcuts import render, redirect
 from .forms import *
 from .models import *
+from django.apps import apps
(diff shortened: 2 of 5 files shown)


app/generate_frontend/generate_landing_page.py
```diff
@@ -0,0 +1,9 @@
+from app.utils import render_template
+
+
+def generate_landing_page_html() -> str:
+    return render_template("landing_page.html.j2", {})
```
NtapSlur, line 5: For rendering the landing_page.html I see that you don't pass any context to the Jinja so that I assume that it is a static web? Did you consider to make a static txt file instead for the rendering since it doesn't inject any value?

app/generate_frontend/generate_landing_page.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def generate_landing_page_views() -> str:
+    return render_template("landing_page_views.py.j2", {})
```
NtapSlur, line 9: Same question as the previous one. I think you can simply like copy the file if it's static and need no value to be inject

app/generate_frontend/generate_landing_page.py
DaWanAnOnli: I still prefer it this way, as there may be modifications in the future that would require the use of context here, for example making the page more pretty or adding more information in the page. As you can see from my template file, it is still quite basic, and does not contain much information.

app/generate_frontend/generate_landing_page.py
DaWanAnOnli: I still prefer it this way, as there may be modifications in the future that would require the use of context here, for example making the page more pretty or adding more information in the page. As you can see from my template file, it is still quite basic, and does not contain much information.

PR #77, opened 2025-04-01
Extra 4 refactor parse class

KenKomKom:
On this PR I've managed to lower the cognitive complexity EVEN FURTHER BEYOND!!!

On the app\parse_json_to_object_class.py I have extracted the methods in parse_class to lower its CC and changed the implementation of parse_relationship and __validate_amount into using design patterns B) as you can see on the new folder parse_class_pattern.

Happy holidays! Cheers [🍺](https://emojipedia.org/beer-mug)!!


NtapSlur:
I love that you refactor the parse relationship algorithm using Strategy Pattern since it makes the code more maintainable and clean! It is nice that finally the CC for Class Diagram Parser is reduced. It seems that the refactor that you made causes a single test not to pass. Is it intentional or maybe there is some logic that changes because of the refactor you made? If it is intentional, it is better to change the test to support the changes you made.

NtapSlur (approved):

PR #78, opened 2025-04-01
PBI 4-8: Integrate Views.py from the parsed with the generated FrontEnd

NtapSlur:
- Integrated views.py from the parsed UML methods with the generated frontend views like CREATE, READ, EDIT, DELETE, and also Landing Page
- Fix the Jinja template for read page views since there is syntax error and causes the generated value to be empty string
- Remove import template in the edit page views since there is already base views template for importing stuff


app/config.py
```diff
@@ -1,7 +1,7 @@
 import os
 
 DEBUG = not os.getenv("PRODUCTION", False)
-VERSION = "0.1.6"
+VERSION = "0.1.7"
```
KenKomKom, line 4: Is this really neccessary?

app/main.py
```diff
@@ -342,9 +347,32 @@ def fetch_data(filename: list[str], content: list[list[str]]) -> dict[str]:
                     )
         for class_method_object in duplicate_class_method_checker.values():
             writer_views.add_class_method(class_method_object)
+
+        # Render the base import
+        response_content_views += render_template("base_views.py.j2", {})
```
KenKomKom, line 352: I suggest changing this to how we usually write strings like how @DaWanAnOnli has explained. Change this to using StringIO. Probably into something like this
res_views = StringIO()
res.write("\n\n")
res.write(writer_views.print_django_style())
res.write(generate_landing_page_views())
res.write(generate_create_page_views())
res.write(generate_read_page_views())
res.write(generate_delete_page_views())
res.write(generate_edit_page_views())

NtapSlur: Thankyou for the suggestion! Gonna change it soon!

KingOnyxx (changes requested):
I think theres a comment you should remove before you merge. Other than that, i approve to merge to dev

tests/test_edit_page_views.py
```diff
@@ -102,5 +102,9 @@ def test_generate_edit_page_views_all_class_empty_fields(self):
...
         self.assertNotIn("def edit_empty_class_one(request, id)", result)
         self.assertNotIn("def edit_empty_class_two(request, id)", result)
-        self.assertIn("from .models import *", result)
-        self.assertIn("from .forms import *", result)
+
+        # Remove assertIn tests since there exists base_views template for
+        # the import & there is no need to import directly from the edit template
+
+        # self.assertIn("from .models import *", result)
+        # self.assertIn("from .forms import *", result)
```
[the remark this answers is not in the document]
NtapSlur: Okay, gonna delete it!

PR #81, opened 2025-04-05
PBI-4-8-Integrate_viewspy_with_formspy_functions

NtapSlur:
Fixed lots of things for the views.py:
- Fixed test_main mock problem to support fetch_data function
- Fixed datatype in method return type to support Python's syntax
- Implemented fetch_data in convert method
- Fixed naming issue for method
- Add import JsonResponse

### DIFF
--- a/app/models/methods.py
+++ b/app/models/methods.py
@@ -154,6 +154,12 @@ class ClassMethodObject(AbstractMethodObject):
                     res.write(" -> bool")
                 elif rettype.lower() == "integer":
                     res.write(" -> int")
+                elif rettype.lower() == "list[string]":
+                    res.write(" -> list[str]")
+                elif rettype.lower() == "list[boolean]":
+                    res.write(" -> list[bool]")
+                elif rettype.lower() == "list[integer]":
+                    res.write(" -> list[int]")
                 else:
                     res.write(f" -> {rettype}")
 


app/main.py
```diff
@@ -78,69 +78,70 @@ async def convert(
...
         )
 
         await download_file(
             request=DownloadRequest(
                 filename=request.filename[0],
-                content=response_content_views,
+                content=render_views(fetched),
                 type="_views",
             ),
         )
```
KingOnyxx, line 147: i think this block of code is not needed anymore because we are sending a ZIP file instead of models and views.
and i also think the test needs to be updated

app/main.py
```diff
@@ -306,81 +307,76 @@ def fetch_data(filename: list[str], content: list[list[str]]) -> dict[str]:
...
-        }
+    # Render the edit views
+    response_content_views.write(generate_edit_page_views(writer_models))
 
-    except ValueError as ex:
-        raise HTTPException(status_code=422, detail=str(ex))
+    return {
+        "models": response_content_models.getvalue(),
+        "views": response_content_views.getvalue(),
+    }
```
KingOnyxx, line 379: can you please add ModelsElements to the return value because our other implementation need to use the object

KingOnyxx (changes requested):
There are a few suggestion of changes that i think you should do before merging

app/main.py
NtapSlur: Gonna add it !

app/main.py
NtapSlur: I think if we delete the download_file it will causes many errors for the main.py test, especially for the endpoint test. Can't we just leave the code or it is a must to delete the download_file?

app/main.py
```diff
@@ -78,69 +78,70 @@ async def convert(
...
+        #         for controller_method_object in controller_method_objects:
+        #             writer_views.add_controller_method(controller_method_object)
+        #         for class_object in class_objects:
+        #             duplicate_class_method_checker = check_duplicate(
+        #                 class_objects, class_object, duplicate_class_method_checker
+        #             )
+        # for class_method_object in duplicate_class_method_checker.values():
+        #     writer_views.add_class_method(class_method_object)
+        # response_content_views += writer_views.print_django_style()
+        # response_content_models += writer_models.print_django_style()
```
MightyZanark, line 131: Just remove this if its already not needed. If you want it just in case, you can always just go back in the commit history.

MightyZanark (changes requested):
Overall looks good, just remove commented code if its already tracked from the previous commit history. If we ever need to go back we can just grab it from the history.

app/main.py
NtapSlur: Okay, gonna delete it!

app/models/methods.py
```diff
@@ -147,7 +147,22 @@ def method_name(request, instance, param1, param2) -> str:
...
+                elif rettype.lower() == "integer":
+                    res.write(" -> int")
+                elif rettype.lower() == "list[string]":
+                    res.write(" -> list[str]")
+                elif rettype.lower() == "list[boolean]":
+                    res.write(" -> list[bool]")
+                elif rettype.lower() == "list[integer]":
+                    res.write(" -> list[int]")
+                else:
+                    res.write(f" -> {rettype}")
```
DaWanAnOnli, line 164: Make sure to change this afterwards, I'm sure there is a more elegant way to write this out. Using multiple elifs like this can cause severe unreadability to the code, causing the code to be very hard to maintain; maintanability issue

app/templates/create_page_views.py.j2
```diff
@@ -4,7 +4,7 @@ def create_{{ class.name_snake }}(request):
         form = {{ class.name }}Form(request.POST)
         if form.is_valid():
             form.save()
-            return redirect('view_{{ class.name_snake }}')
+            return redirect('get_{{ class.name_snake }}')
```
DaWanAnOnli, line 7: there is still a bug here. it needs to be 'main:get_{{ ...' We will fix this and similar instances in the PBI 4 branch

DaWanAnOnli (commented):
All good, just a few notes regarding what needs to be fixed after merging to PBI 4

PR #106, opened 2025-04-11
Monitoring

KenKomKom:
In this PR I implemented prometheus custom metrics to :
1. Count the number of errors depending on what type they are
2. Check the duration of parsing both class and sequence diagram

If you have anymore suggestions on what we should monitor, let me know [🍺](https://emojipedia.org/beer-mug)


KenKomKom:
> Does this already show up on prometheus? Since, when I added prometheus here, I used the `prometheus_fastapi_instrumentator` library to expose the metrics and I don't know if it will automatically add the metrics.
> 
> If it doesn't appear, I think you might want to take a look at [this](https://github.com/trallnag/prometheus-fastapi-instrumentator/tree/master?tab=readme-ov-file#creating-new-metrics) for adding the metrics to the instrumentator.

Yup, it can be queried after there is data to be queried. I already run it using docker-compose and it worked

PR #108, opened 2025-04-13
Extra 2 2 create jinja for views.py

DaWanAnOnli:
- Create Jinja implementation for views.py functions from sequence diagram.

Comment deadline: 14/4/2025 15.00 WIB


app/templates/sequence_views.py.j2
```diff
@@ -0,0 +1,19 @@
+{% for class_method in class_methods -%}
+#-----method from class {{ class_method.class_name }} ------
+
+def {{ class_method.method_name }}(request, instance_name{% for param in class_method.params %}, {{ param.param_name }}: {{ param.param_type }}{% endfor %}){% if class_method.return_type != "void" %} -> {{ class_method.return_type }}{% endif %}:
+    {% for method_call in class_method.method_calls %}{% if method_call.condition != "" %}if {{ method_call.condition }}:
+        {% endif %}{% if method_call.return_var_name != "" %}{{ method_call.return_var_name }} = {% endif %}{{ method_call.method_name }}(request, {{ method_call.instance_name }}{% for argument in method_call.arguments %}, {{ argument.argument_name }}{% endfor %})
```
MightyZanark, line 6: Rather than checking if a value is an empty string, it seems you can just check if the value is defined instead: https://jinja.palletsprojects.com/en/stable/templates/#jinja-tests.defined.

app/models/methods.py
```diff
@@ -344,6 +402,27 @@ def print_django_style(self) -> str:
...
 
+    def print_django_style_template(self) -> dict[str]:
+        context = {}
+
+        context["condition"] = ""
+        if self.__condition:
+            context["condition"] = self.__condition
+        context["return_var_name"] = ""
+        if self.__return_var_name:
+            context["return_var_name"] = self.__return_var_name
```
MightyZanark, line 413: If there is no condition nor return value, you can just not put anything in the context for it. On the template side, you can check if a variable is defined or not like this https://jinja.palletsprojects.com/en/stable/templates/#jinja-tests.defined.

MightyZanark (changes requested):
Overall looks good! Thanks for adding this part as last time around I didn't manage to add the views version. Just have some small change I would like you to do.

KingOnyxx (changes requested):
Thank you for the updates in this PR! After reviewing, there is a few changes from me and @MightyZanark that i think you need to implement before merging this PR. Other than that, i think it looks good to me!

app/models/elements.py
```diff
@@ -154,6 +154,22 @@ def add_class_method(self, class_method_object: ClassMethodObject):
...
+        ]
+        controller_method_context = [
+            controller_method_object.print_django_style_template()
+            for controller_method_object in self.__controller_methods
+        ]
+
+        context["class_methods"] = class_method_context
+        context["controller_methods"] = controller_method_context
+
+        return render_template("sequence_views.py.j2", context)
```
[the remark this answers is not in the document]
DaWanAnOnli: done

app/templates/sequence_views.py.j2
```diff
@@ -0,0 +1,19 @@
...
+    {% endfor %}# TODO: Auto generated function stub
+    raise NotImplementedError('{name} function is not yet implemented')
```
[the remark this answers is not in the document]
DaWanAnOnli: Done

app/templates/sequence_views.py.j2
DaWanAnOnli: Done

app/models/methods.py
DaWanAnOnli: Done

app/templates/sequence_views.py.j2
```diff
@@ -0,0 +1,19 @@
...
+    pass
+
+
+{% endfor %}
+{%- for controller_method in controller_methods %}
```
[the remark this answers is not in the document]
DaWanAnOnli: Good point. However I believe this is not critical for now. However this is something that should be addressed later.

PR #112, opened 2025-04-17
Pbi 9 2 setup value injection for post request in swagger

KenKomKom:
Continuing the other PR, on this PR I've implemented the essentials for controller.java to use swagger. So, there will be a function for getting the necessary imports and the decorators.

Cheers! [🍺](https://emojipedia.org/beer-mug)

### DIFF
--- a/app/generate_swagger/generate_swagger.py
+++ b/app/generate_swagger/generate_swagger.py
@@ -1,0 +1,7 @@
+from app.utils import render_template
+
+
+# Call this to generate SwaggerConfig.java
+def generate_swagger_config(project_name: str) -> str:
+    context = {"project_name": project_name}
+    return render_template("SwaggerConfig.java.j2", context)
--- a/app/generate_swagger/generate_swagger.py
+++ b/app/generate_swagger/generate_swagger.py
@@ -1,3 +1,4 @@
+from app.models.diagram import ClassObject
 from app.utils import render_template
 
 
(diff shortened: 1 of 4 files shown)


MightyZanark (approved):
Looks like a small addition as well, but there seems to be commits already on another PR. I don't remember quite exactly how it might merge but it might create duplicate commits in the history if you merge the other one before this one. But again I might be remembering wrong how merging with merge commits work. Overall, approved!

app/generate_swagger/generate_swagger.py
```diff
@@ -0,0 +1,24 @@
...
+# Call this to generate SwaggerConfig.java
+def generate_swagger_config(project_name: str) -> str:
+    context = {"project_name": project_name}
+    return render_template("SwaggerConfig.java.j2", context)
+
+
+# I made this function just in case when integration we forgot to
+# add the swagger import to controller, I hope be this function will remind us of it
+def get_swagger_controller_import() -> str:
+    return "import io.swagger.v3.oas.annotations.Operation;\n"
```
DaWanAnOnli, line 14: I believe this function is still unecessary, even with the comments that you have given. This is due to the fact that the correct controller file example is already available at the generated-burhanpedia repository. This way template generation for controller files will be much easier, and all the import statements have already been put in the example project.

I'm afraid that this will bring confusion in the future, as we wonder whether this function is ever used or not. The worst case will be if someone uses this function for their template, while others don't. This is definitely a maintenance issue. To avoid this, I believe it will be better to simply remove it, and put the import statement directly in the Jinja template.

DaWanAnOnli (commented):
A small comment regarding get_swagger_controller_import function

PR #113, opened 2025-04-17
Pbi 7 3 generate repository file

KenKomKom:
On this PR I've implemented a utils function to turn snake case to pascal case for class names. Also I've also implemented file generation for repository files in springboot.

Cheers! [🍻](https://emojiterra.com/clinking-beer-mugs/)

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -47,6 +47,26 @@ def camel_to_snake(camel_case_str: str) -> str:
     return snake_case_str.lower()
 
 
+def to_pascal_case(s: str, acronyms: list[str] = None) -> str:
+    if acronyms is None:
+        acronyms = {"API", "HTTP", "XML", "ID", "URL", "JSON"}  # Add more as needed
+
+    # Normalize delimiters
+    s = re.sub(r"[-_]", " ", s)
+
+    words = s.split()
+    result = []
+
+    for word in words:
+        upper_word = word.upper()
+        if upper_word in acronyms:
+            result.append(upper_word)
+        else:
+            result.append(word.capitalize())
+
+    return "".join(result)
+
+
 def render_project_django_template(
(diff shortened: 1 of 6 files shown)


app/utils.py
```diff
@@ -47,6 +47,26 @@ def camel_to_snake(camel_case_str: str) -> str:
     return snake_case_str.lower()
 
 
+def to_pascal_case(s: str, acronyms: list[str] = None) -> str:
+    if acronyms is None:
+        acronyms = {"API", "HTTP", "XML", "ID", "URL", "JSON"}  # Add more as needed
```
MightyZanark, line 52: The `acronyms` param type hint doesn't seem to match what you give it as a default value. It should be `set[str]` (or better yet `Optional[set[str]]`). Also, will handling these specific cases worth the effort compared to just treating all of them the same way (Api, Http, Xml, etc)?

tests/test_generate_repository.py
```diff
@@ -0,0 +1,18 @@
...
+import unittest
+
+from app.generate_repository.generate_repository import generate_repository_java
+from app.models.diagram import ClassObject
+
+CUR_DIR = os.path.dirname(os.path.realpath(__file__))
+TEST_DIR = os.path.join(CUR_DIR, "testdata")
+
+
+# Tests for generate_html_read_page_django
```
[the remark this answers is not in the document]
MightyZanark: Yea, seems like you forgot to change the comment.

MightyZanark (changes requested):
There are some changes I would like you to do considering the result of our daily meet discussions yesterday and some other slightly minor stuff.

app/utils.py
KenKomKom: Yea.. they're not necessary, but I just think it'll be nicer you know

PR #114, opened 2025-04-17
Pbi 7 5 set up sqlite for springboot

KenKomKom:
On this PR, I have implemented write for SQLiteDialect for Springboot using jinjia2 as its base template for injecting project name into the file. I assume the configs on application.properties and build.gradle.kts will be implemented on another task.

Cheers! [🍻](https://emojiterra.com/clinking-beer-mugs/)

### DIFF
--- a/app/templates/SQLiteDialect.java.j2
+++ b/app/templates/SQLiteDialect.java.j2
@@ -1,4 +1,4 @@
-package com.example.{{project_name}}.config;
+package {{trailing_project_name}}.{{project_name}}.config;
 
 import org.hibernate.dialect.Dialect;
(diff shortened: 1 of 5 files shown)


app/templates/SQLiteDialect.java.j2
```diff
@@ -0,0 +1,81 @@
+package com.example.{{project_name}}.config;
```
MightyZanark, line 1: Similar comment as the other PR, the `com.example` part should be changed to be extracted from context instead now.

MightyZanark (changes requested):
Overall looks good, just a similar comment as the other PR. Other than that, approved!

PR #115, opened 2025-04-18
PBI-7-2: Generate Service for Simple CRUD

NtapSlur:
- Added template for Service.java and create method to render based on the project_name and ClassObject
- Implemented BDD also using pytest-bdd and added pytest-bdd to the requirements-tests.txt

Important note:
In previous PR of Ken, Max gives suggestion to change the package name so that it passes the value from context. But for the jinja template for service, I'll temporarily use the com.example first as the package name since it needs to modify the main.py first to pass the value of the package name from the request payload.  Gonna implement the package name as context on the integration!


app/templates/springboot/service.java.j2
```diff
@@ -0,0 +1,40 @@
...
+@Service
+@RequiredArgsConstructor
+public class {{class_name_capital}}Service {
+
+    private final {{class_name_capital}}Repository {{class_name}}Repository;
+
+    // Method for create {{class_name}} instance
+    public Cart create{{class_name_capital}}({{class_name_capital}} {{class_name}}) {
+        return {{class_name}}Repository.save({{class_name}});
+    }
```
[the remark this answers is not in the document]
NtapSlur: Seems forgot to change that. Gonna change it!

app/templates/springboot/service.java.j2
```diff
@@ -0,0 +1,40 @@
...
+    public {{class_name_capital}} get{{class_name_capital}}(String id) {
+        return {{class_name}}Repository.findById(id).orElse(null);  // Return null if not found
+    }
+
+    // Method to update {{class_name}} instance by id
+    public {{class_name_capital}} update{{class_name_capital}}({{class_name_capital}} {{class_name}}, String id) {
+        {{class_name_capital}} existing{{class_name_capital}} = {{class_name}}Repository.findById(id).orElse(null);
+
+        if (existing{{class_name_capital}} != null) {
+            existing{{class_name_capital}}.setFull({{class_name}}.isFull());
```
DaWanAnOnli, line 29: This is a code specific to a mock class I made in burhanpedia project. So this needs to be generalized. 

The way to generalize it is to loop for fields in the class. Then you need to set the each field of the existing class to the new field in the new class. 

In the current implementation, you are not updating the object at all. Instead, it will most likely give an error as the object does not have the method .setFull.

This needs to be fixed.

DaWanAnOnli (changes requested):
A bug in the Jinja template regarding the update method in service file. Needs to be fixed.

app/templates/springboot/service.java.j2
```diff
@@ -0,0 +1,40 @@
+package com.example.{{project_name}}.service;
+
+import com.example.{{project_name}}.model.{{class_name_capital}};
+import com.example.{{project_name}}.repository.{{class_name_capital}}Repository;
```
KingOnyxx, line 4: are you sure to force package name to start with com.example? i think its better to discuss this with the team first.

app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -0,0 +1,24 @@
...
+    class_name_capital, class_name = format_class_name(model.get_name())
+    context = {
+        "project_name": project_name,
+        "class_name_capital": class_name_capital,
+        "class_name": class_name,
+    }
+    return render_template("springboot/service.java.j2", context)
+
+def format_class_name(name: str) -> tuple[str, str]:
+    return name.capitalize(), name[0].lower() + name[1:]
```
KingOnyxx, line 24: what happens if the class object name is more than 1 word for example Payment Gateway?

KingOnyxx (changes requested):
After reviewing, i have some concerns regarding your implementation in this PR. Please consider fixing before merging this PR

app/templates/springboot/service.java.j2
NtapSlur: Oops my bad! I thought that it was a method from Springboot to insert value to the repository. Gonna fix it soon!

app/templates/springboot/service.java.j2
NtapSlur: I think i've explained this in the description. For now, I'm waiting for the integration to inject the correct value for the package name

app/generate_service_springboot/generate_service_springboot.py
NtapSlur: Isn't that already handled in the parse class in class naming convention for MoTxT? Or is the case for more than 1 word not handled yet? Cmiiw*

KenKomKom: get ready to change the class_name_capital to use `to_pascal_case(class_name)` from the PBI-7-3 or 5 (I forgot) to improve consistency in the generated code and remove snake case.

PR #118, opened 2025-04-20
PBI 8: 3 - feat: function to initialize spring boot project from self hosted spring initializr

MightyZanark:
Part of MOT-108
- Added function to initiate zip file creation of the spring boot project through spring initializr. This function will return the temp file path. This function assumes that the `project_name` and `group_id` is already validated from the caller.
- Changed the `update_version` workflow so now if there are pull request from any branch with the pattern `BUG-FIX-*`, we can make the target branch staging or main to quickly deliver it to the deployed build, rather than having to go through dev.


KingOnyxx (approved):
Thank you for implementing these changes! After reviewing it, i didn't find anything thats concerning to me. Approved!

PR #119, opened 2025-04-20
Create to camel case function

DaWanAnOnli:
This PR implements the to_camel_case function that converts every string into camel case.

This function might seem simple, but it requires a lot of tests and experiments as there are many edge cases.

This PR might seem small however it is important as this function will be heavily used by other branches in this PBI. Therefore this function must be merged immediately.

Deadline Comment: 21 April 2025 11.00 WIB.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -48,9 +48,6 @@ def camel_to_snake(camel_case_str: str) -> str:
 
 
 def to_camel_case(s: str) -> str:
-    if not s:
-        return ""
-
     # Remove non-alphanumeric characters, replace with spaces, and split by spaces
     words = re.sub(r"[^a-zA-Z0-9]", " ", s).split()
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -56,9 +56,9 @@ def to_camel_case(s: str) -> str:
 
     # Check the first word and ensure it's lowercase if it's not already camelCase
     first_word = words[0]
-    if first_word[
-        0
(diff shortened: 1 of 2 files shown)


app/utils.py
```diff
@@ -47,6 +47,34 @@ def camel_to_snake(camel_case_str: str) -> str:
     return snake_case_str.lower()
 
 
+def to_pascal_case(string_input: str) -> str:
+    return ""
```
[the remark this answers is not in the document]
DaWanAnOnli: Fixed

app/utils.py
```diff
@@ -47,6 +47,27 @@ def camel_to_snake(camel_case_str: str) -> str:
...
+    words = re.sub(r"[^a-zA-Z0-9]", " ", s).split()
+
+    if not words:
+        return ""
+
+    # Check the first word and ensure it's lowercase if it's not already camelCase
+    first_word = words[0]
+    if first_word[
+        0
+    ].isupper():  # If the first character is uppercase, make it lowercase
```
MightyZanark, line 61: Is this formatting from the linter? This just makes it kinda weird to read if its not from the linter, but if it is, I think it should be ignored.

MightyZanark (changes requested):
Looks good, don't have much to say since it is also pretty straightforward. Only remark I have is below.

app/utils.py
DaWanAnOnli: Done

PR #120, opened 2025-04-21
PBI-7-6-Generate Service Method From Diagram

NtapSlur:
Added new feature for generate_service so that it can get the methods from Class Diagram:
- Modify class diagram parser to keep the method modifier (+/-) and not just replace it with \n
- Modify AbstractMethodObject to include modifier as an attribute
- Added to_springboot_code in ClassMethodObject
- Added is_public logic for endpoint


DaWanAnOnli:
Wrong branch to merge to. It should be PBI 7, not main

DaWanAnOnli (commented):
Comment regarding to_springboot_code method

app/models/properties.py
```diff
@@ -116,6 +116,14 @@ class ParameterObject:
...
     """
 
+    JAVA_TYPE_MAPPING = {
+        "boolean": "boolean",
+        "bool": "boolean",
+        "string": "String",
+        "str": "String",
+        "integer": "int",
+    }
+
```
KingOnyxx, line 126: The `JAVA_TYPE_MAPPING` dictionary assumes a specific mapping Java types. Please consider using my implementation in PBI-7-1 branch for more mappings.

app/models/methods.py
```diff
@@ -178,6 +193,61 @@ def method_name(request, instance, param1, param2) -> str:
...
 
+    def to_springboot_code(self) -> str:
+        """
+        Return Springboot representation of method in the UML Diagram
+
+        This implementation ignore method calls so only supports for Class Diagram
+        """
+        name = self.get_name()
+
+        if name is None or not is_valid_python_identifier(name):
```
KingOnyxx, line 204: i think its not valid to check Java name with python identifier

KingOnyxx (changes requested):
Thank you for implementing these changes! I have some suggestions that you consider changing before merging. Other than that, these changes look good to me.

app/models/methods.py
```diff
@@ -178,6 +193,61 @@ def method_name(request, instance, param1, param2) -> str:
 
         return res.getvalue()
 
+    def to_springboot_code(self) -> str:
+        """
+        Return Springboot representation of method in the UML Diagram
+
+        This implementation ignore method calls so only supports for Class Diagram
+        """
```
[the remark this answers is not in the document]
NtapSlur: Of course we can! Kinda agree with that so that if there are changes in the future, it won't be that hard to modify the template. Gonna change it!

app/models/properties.py
NtapSlur: I actually just follow on what is accepted based on the user manual.  I see that you handle many data types, which is great! Just curious, for the datatype like UUID, LocalDate, etc, doesn't it need to be imported? But for the rest like float and double, i can add it!

app/models/methods.py
NtapSlur: I think for now it is fine (although it doesn't follow Java's convention name for Lower Camel Case), but Python's isidentifier checks some rules that are similiar like no number in the beginning of the variable name, only allow underscore and alphanumeric. If there are any edge cases that should be allowed in Java but not in Python i think we can consider in making our own is_valid_java_identifier.

PR #121, opened 2025-04-21
PBI 8: 8 - Make Spring Project Downloadable

MightyZanark:
Part of MOT-123
- Modified the `convert` function so that it now can accept either `django` (default) or `spring` project types. When the project type is `spring`, if the `group_id` is not sent by the user, then it will be defaulted to `com.example` for now.
- Added logging to unexpected exceptions

The function I added in this PR does not do much currently, but when we are going to integrate, the integrators can just put it in the provided function rather than having to modify a lot more. Currently the function just checks for the `group_id` if it is a valid Java package name or not. If it is then it just create some temporary file, which if PBI 8-3 is merged, it should be replaced by that function instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.11'
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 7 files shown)


KingOnyxx (approved):
Thanks for implementing these changes! After reviewing, i don't think that there's any problem for me. Approved!

PR #123, opened 2025-04-21
Extra 6 refactor create django project

KingOnyxx:
Part of MOT-128
This pull request refactors the Django project generation process to improve efficiency and simplify the code. The key changes include replacing folder-based file handling with in-memory operations using `zipfile.writestr`, modifying the `render_project_django_template` function to return a dictionary instead of creating files on disk, and updating tests to reflect these changes.

### Refactoring of Django project generation:

* [`app/main.py`](diffhunk://#diff-990f7e4140fab1ad82afc787505090b64d4024e63a891405cd9085e7e6b249ddL219-R228): Replaced manual file writing and zipping with an in-memory approach using `zipfile.ZipFile.writestr`. This eliminates the need to create temporary folders on disk.
* [`app/utils.py`](diffhunk://#diff-ac7b8d56e509c921efaadb5a776e1c9037531c4d2af78787f06f67a1f3533ae4L52-L60): Updated `render_project_django_template` to return a dictionary of file names and their contents instead of creating files in a folder. This change supports the in-memory zipping process. [[1]](diffhunk://#diff-ac7b8d56e509c921efaadb5a776e1c9037531c4d2af78787f06f67a1f3533ae4L52-L60) [[2]](diffhunk://#diff-ac7b8d56e509c921efaadb5a776e1c9037531c4d2af78787f06f67a1f3533ae4L69-R65)

### Updates to test cases:

* [`tests/test_render_project_django.py`](diffhunk://#diff-eaf3b5974ce3b0b89a1dc77c8f57d3daea94ec9c2b75962b7e2a0d4c4d3c6f7bL36-R68): Updated `test_generate_django_project_positive` to validate the contents of the in-memory zip file instead of checking for folder existence. Adjusted file comparison logic to match the new approach.
* Removed tests that relied on folder creation (`test_render_project_django_template_positive` and `test_render_project_django_template_negative_with_existing_folder`) as they are no longer relevant with the new in-memory implementation. [[1]](diffhunk://#diff-eaf3b5974ce3b0b89a1dc77c8f57d3daea94ec9c2b75962b7e2a0d4c4d3c6f7bL74-L87) [[2]](diffhunk://#diff-eaf3b5974ce3b0b89a1dc77c8f57d3daea94ec9c2b75962b7e2a0d4c4d3c6f7bL100-L115)

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -49,15 +49,10 @@ def camel_to_snake(camel_case_str: str) -> str:
 
 def render_project_django_template(
     template_path: str, context: dict[str, Any]
-) -> list[str]:
-    files = []
+) -> dict[str, str]:
+    files = {}
     if not is_valid_python_identifier(context["project_name"]):
         raise ValueError("Project name must not contain whitespace or number!")
-    folder_path = f"project_{context['project_name']}"
-    try:
-        os.makedirs(folder_path)
-    except OSError:
-        raise FileExistsError(f"Folder {folder_path} already exists")
     for template_name in os.listdir(template_path):
         template = f"django_project/{template_name}"
         if template_name == "settings.py.j2":
@@ -66,10 +61,8 @@ def render_project_django_template(
                 "SECRET_KEY": get_random_secret_key(),
             }
         template_name = template_name.replace(".j2", "")
-        with open(os.path.join(folder_path, template_name), "w") as file:
-            file.write(render_template(template, context))
-            file.write("\n")  # add newline at the end of file for linter
-        files.append(template_name)
+        file = render_template(template, context) + "\n"
+        files[template_name] = file
     return files
 
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -51,18 +51,18 @@ def render_project_django_template(
     template_path: str, context: dict[str, Any]
 ) -> dict[str, str]:
     files = {}
-    if not is_valid_python_identifier(context["project_name"]):
+    project_name = context["project_name"]
+    if not is_valid_python_identifier(project_name):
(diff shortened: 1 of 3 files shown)


KenKomKom:
I believe the changes are done well to improve the write process. LGTM

app/utils.py
```diff
@@ -66,10 +61,8 @@ def render_project_django_template(
                 "SECRET_KEY": get_random_secret_key(),
             }
         template_name = template_name.replace(".j2", "")
-        with open(os.path.join(folder_path, template_name), "w") as file:
-            file.write(render_template(template, context))
-            file.write("\n")  # add newline at the end of file for linter
-        files.append(template_name)
+        file = render_template(template, context) + "\n"
+        files[template_name] = file
```
MightyZanark, line 65: I have some concerns that may or may not happen regarding saving it in memory before writing it to the temp zip. In the worse case, if someone made a valid diagram such that it has a lot of classes with attributes and the likes, if we put the rendered result all in memory right here, we might encounter an Out of Memory error. What I think would be better is make it so that the value of the dict is a lambda function that when called, will then execute the `render_template(template, context)` code. This way, we only unload the template when we are going to write it in the zip.

MightyZanark (changes requested):
Overall looks good. Thanks for making it so that it now doesn't make arbitrary folders in the current directory. I just have some small suggestion that might reduce the possibility of an Out of Memory error.

app/utils.py
KingOnyxx: Thanks for noticing! I'll fix it ASAP

PR #125, opened 2025-04-23
BUG-FIX: Fix Generated Edit for Service

NtapSlur:
- Now, the edit method will iterate all available FieldObject (class' attributes) to call the setter and getter from lombok
- Refactor JAVA_TYPE_MAPPING to be as constant in utils.py so that if needed in future, we only need to import the constant

### DIFF
--- a/tests/springboot/test_generate_service.py
+++ b/tests/springboot/test_generate_service.py
@@ -91,8 +91,6 @@ def render_template_output(context):
 def check_output(context):
     with open("tests/springboot/test_service_data.txt", "r", encoding="utf-8") as file:
         expected_output = file.read()
-    
-    print(context["output"])
 
     assert context["output"].replace(" ", "").replace("\n", "").strip() == expected_output.replace(" ", "").replace("\n", "").strip()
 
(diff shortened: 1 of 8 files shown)


tests/springboot/test_generate_service.py
```diff
@@ -24,6 +24,18 @@ def test_generate_service(self):
...
+        field.set_name("isFull")
+        field_type.set_name("boolean")
+        field.set_type(field_type)
+        class_object.add_field(field)
+        field = FieldObject()
+        field_type = TypeObject()
+        field.set_name("cartId")
+        field_type.set_name("int")
+        field.set_type(field_type)
+        class_object.add_field(field)
```
KingOnyxx, line 38: i think its better to use setUp() and put these code inside the function to avoid code duplication

KingOnyxx (changes requested):
Thank you for implementing these changes! After reviewing it, i have found a problem in your tests. Please consider to change it before merging. Thanks

tests/springboot/test_generate_service.py
NtapSlur: For the object instantiation is only for test_generate_service test only because for the BDD test is already set up using prepare_context() so I think it is okay not to use setUp()

DaWanAnOnli (approved):
Approved by me

PR #126, opened 2025-04-24
Pbi 8 6 determine and add spring dependencies (application.properties)

KingOnyxx:
Part of MOT-113
This pull request introduces a new method to generate Spring Boot `application.properties` content and adds corresponding unit tests to ensure its correctness and handle edge cases. The most important changes include the addition of the `print_application_properties` method in the `DependencyElements` class and new test cases in the `tests/test_elements.py` file.

### Enhancements to `DependencyElements`:

* Added `print_application_properties` method to generate Spring Boot `application.properties` content. It includes default configurations and returns a string. The method raises a `ValueError` if the provided project name is empty. (`app/models/elements.py`)

### DIFF
--- a/app/models/elements.py
+++ b/app/models/elements.py
@@ -275,6 +275,9 @@ class DependencyElements(FileElements):
     def print_django_style(self) -> str:
         return super().print_django_style()
 
+    def print_application_properties(self, project_name: str) -> str:
+        return ""
+
     def print_springboot_style(self, project_name: str) -> str:
         context = {
             "project_name": project_name,
--- a/app/models/elements.py
+++ b/app/models/elements.py
@@ -272,7 +272,10 @@ class DependencyElements(FileElements):
     This class is only for writing script for dependency in springboot framework
     """
 
-    def print_django_style(self) -> str:
+    def print_django_style(self) -> str:  # pragma: no cover
+        """
+        Only for abstract method purposes and doesn't return anything
+        """
         return super().print_django_style()
 
     def print_application_properties(self, project_name: str) -> str:
--- a/app/models/elements.py
+++ b/app/models/elements.py
@@ -276,7 +276,17 @@ class DependencyElements(FileElements):
         return super().print_django_style()
 
     def print_application_properties(self, project_name: str) -> str:
-        return ""
+        if project_name == "":
+            raise ValueError("Project name cannot be empty!")
+        return (
+            "springdoc.api-docs.enabled=true\n"
+            + "springdoc.swagger-ui.enabled=true\n"
+            + "spring.datasource.url=jdbc:sqlite:mydatabase.db\n"
+            + "spring.datasource.driver-class-name=org.sqlite.JDBC\n"
+            + "spring.jpa.show-sql=true\n"
+            + "spring.jpa.database-platform=org.hibernate.community.dialect.SQLiteDialect\n"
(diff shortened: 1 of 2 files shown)


app/models/elements.py
```diff
@@ -272,9 +272,25 @@ class DependencyElements(FileElements):
...
+        """
+        Only for abstract method purposes and doesn't return anything
+        """
         return super().print_django_style()
 
+    def print_application_properties(self, project_name: str) -> str:
+        if project_name == "":
+            raise ValueError("Project name cannot be empty!")
+        return (
+            "springdoc.api-docs.enabled=true\n"
```
KenKomKom, line 285: I think it's better to move it to a file for better readability and maintainability in the future. Then for this, just read from the file and return the content as string

KingOnyxx: In my opinion, application.properties file can be considered a static file and doesn't need to use a templating engine (like Jinja). it’s almost always cleaner to skip the temporary file (`.txt`) and use `writestr()` from ZipFile library in python instead.

app/models/elements.py
```diff
@@ -272,9 +272,25 @@ class DependencyElements(FileElements):
...
+            + "springdoc.swagger-ui.enabled=true\n"
+            + "spring.datasource.url=jdbc:sqlite:mydatabase.db\n"
+            + "spring.datasource.driver-class-name=org.sqlite.JDBC\n"
+            + "spring.jpa.show-sql=true\n"
+            + "spring.jpa.database-platform=org.hibernate.community.dialect.SQLiteDialect\n"
+            + "spring.jpa.hibernate.ddl-auto=update\n"
+        )
```
MightyZanark, line 292: Why does it need a `project_name` parameter when it does not even use it? Also, project name being an empty string should already be handled by the Pydantic model in `model.py`, so we don't really need to check it again.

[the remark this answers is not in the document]
MightyZanark: I agree with Gemini, I think it is better to at least make the configuration extendable if needed. We may not need it now, but if we want to expand and add some more config in the `application.properties` according to the user's input, it would be better if we just have to add onto it.

MightyZanark (changes requested):
Overall looks good, but I have some opinions regarding the design decision you made in creating this function.

app/models/elements.py
```diff
@@ -272,9 +272,25 @@ class DependencyElements(FileElements):
     This class is only for writing script for dependency in springboot framework
     """
 
-    def print_django_style(self) -> str:
+    def print_django_style(self) -> str:  # pragma: no cover
```
DaWanAnOnli, line 275: I still don't understand why this function is here. If it is not here, won't the print_djago_file method from the superclass will still be called? If there is no purpose, then at the very least it will simply take unnecessary space. However, this may lead to confusion for future developers, leading to maintenance issue.

DaWanAnOnli (commented):
A comment regarding print_django_style function in DependencyElements class

app/models/elements.py
KingOnyxx: Documentation from ABC
> A class that has a metaclass derived from ABCMeta cannot be instantiated unless all of its abstract methods are overridden. The abstract methods can be called using any of the normal 'super' call mechanisms.

Since DependencyElements inherits FileElements which inherits ABC and print_django_style() in FileElements is using `@abstractmethod` decorator, hence why i put that there.

PR #128, opened 2025-04-24
Extra 7 1 implement is public parsing from diagram

DaWanAnOnli:
Implemented task EXTRA-7-1

I have added modifier field in FieldObject, and have implemented the parsing of class diagram attributes to set the modifier to public (using "+") or private (using "-")

### DIFF
--- a/app/models/properties.py
+++ b/app/models/properties.py
@@ -141,8 +141,9 @@ class FieldObject:
 
         for key, value in self.SPRING_TYPE_MAPPING.items():
             if key.lower() in field_type:
-                return {"name": self.__name, "type": value}
-        return {"name": self.__name, "type": "String"}  # Default fallback
+                return {"name": self.__name, "type": value, "modifier": self.__modifier}
+        # Default fallback
+        return {"name": self.__name, "type": "String", "modifier": self.__modifier}
 
 
 class ParameterObject:
--- a/app/models/properties.py
+++ b/app/models/properties.py
@@ -94,6 +94,7 @@ class FieldObject:
     def __init__(self):
         self.__name: str = ""
         self.__type: Optional[TypeObject] = None
+        self.__modifier = ""
 
     def __str__(self) -> str:
         return f"FieldObject:\n\tname: {self.__name}\n\ttype: {self.__type}"
@@ -106,6 +107,13 @@ class FieldObject:
 
     def set_type(self, type: TypeObject):
         self.__type = type
+
+    def set_modifier(self, modifier: str):
+        if (modifier != "public") and (modifier != "private"):
+            raise ValueError(
+                f'Class field modifier must be either "public" or "private" ! Got: {modifier}'
+            )
+        self.__modifier = modifier
 
     def to_models_code(self) -> str:
         field_type = self.__type.to_models_code().lower()
--- a/app/models/properties.py
+++ b/app/models/properties.py
@@ -94,7 +94,7 @@ class FieldObject:
     def __init__(self):
         self.__name: str = ""
         self.__type: Optional[TypeObject] = None
-        self.__modifier = ""
+        self.__modifier = "private"
 
     def __str__(self) -> str:
         return f"FieldObject:\n\tname: {self.__name}\n\ttype: {self.__type}"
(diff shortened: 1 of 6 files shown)


app/models/properties.py
```diff
@@ -94,6 +94,7 @@ class FieldObject:
     def __init__(self):
         self.__name: str = ""
         self.__type: Optional[TypeObject] = None
+        self.__modifier = ""
```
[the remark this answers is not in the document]
DaWanAnOnli: Done

tests/test_model_springboot.py
```diff
@@ -94,17 +94,27 @@ def test_to_springboot_models_template_positive(self):
...
+            "type": "Integer",
+            "modifier": "private",
+        }
         self.assertEqual(
             self.field_object.to_springboot_models_template(), expected_output
         )
 
     def test_to_springboot_models_template_default_fallback(self):
         self.field_object.set_name("test_field")
+        self.field_object.set_modifier("private")
```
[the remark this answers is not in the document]
DaWanAnOnli: Done

tests/test_model_springboot.py
```diff
@@ -114,7 +124,8 @@ def test_to_springboot_models_template_edge_case_empty_name(self):
         type_obj = TypeObject()
         type_obj.set_name("boolean")
         self.field_object.set_type(type_obj)
-        expected_output = {"name": "", "type": "boolean"}
+        self.field_object.set_modifier("private")
```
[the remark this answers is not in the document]
DaWanAnOnli: Done

tests/test_model_springboot.py
```diff
@@ -94,17 +94,27 @@ def test_to_springboot_models_template_positive(self):
         type_obj = TypeObject()
         type_obj.set_name("integer")
         self.field_object.set_type(type_obj)
-        expected_output = {"name": "test_field", "type": "Integer"}
+        self.field_object.set_modifier("private")
```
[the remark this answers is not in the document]
DaWanAnOnli: Done

app/models/properties.py
```diff
@@ -94,6 +94,7 @@ class FieldObject:
     def __init__(self):
         self.__name: str = ""
         self.__type: Optional[TypeObject] = None
+        self.__modifier = "private"
```
MightyZanark, line 97: If the value for modifier is only `public` or `private`, I think we can just make it a boolean. But, if we are looking forward for an extension (like `protected` or similar), you can annotate the attribute with the `Literal` type from `typing` as can be seen [here](https://docs.python.org/3/library/typing.html#typing.Literal).

app/models/properties.py
```diff
@@ -107,6 +108,13 @@ def set_name(self, name: str):
     def set_type(self, type: TypeObject):
         self.__type = type
 
+    def set_modifier(self, modifier: str):
+        if (modifier != "public") and (modifier != "private"):
```
NtapSlur, line 112: Just curious, have you considered about the default modifier in Java? Like if the user doesn't use +/- it should be classified as default modifier. On the method modifier it has already handled for default modifier. I think that it is a good thing to also handle default modifier

MightyZanark (approved):
Overall looks good to me! Just a simple question and suggestion for the `modifier` attribute. Other than that, approved!

app/parse_json_to_object_class.py
```diff
@@ -201,10 +201,20 @@ def __create_parameter(self, parameter: str) -> ParameterObject:
...
-
-        attr = FieldObject()
+        attr_object = FieldObject()
         attr_type = TypeObject()
+        modifier = attribute.strip(" ")[0]  # check first character
+
+        if modifier == "+":
+            attr_object.set_modifier("public")
+        elif modifier == "-":
+            attr_object.set_modifier("private")
```
KenKomKom, line 211: I think it's best to also support # which is a valid uml class diagram symbol for a protected attribute

app/models/properties.py
DaWanAnOnli: Regarding default modifier, I think it is worth considering. However, since I believe UML Class Diagram specification requires users to put modifiers no matter what, I don't think it is the most urgent thing\

Regarding protected, I believe it is something we should implement. However, to speed up integration process, we should just stick to this first and make sure everything works first.

app/models/properties.py
DaWanAnOnli: Noted. I think we can consider implementing this in refactoring stage.

app/parse_json_to_object_class.py
DaWanAnOnli: I believe it is something we should implement it. However, to speed up integration process, we should just stick to this first and make sure everything works first.

PR #132, opened 2025-04-25
PBI-8-4-Insert Generated File into Springboot Init

NtapSlur:
- All the files including model, repository, service, controller, and application.properties are inserted into the spring zip

Comment Deadline: 26/04/2025


app/main.py
```diff
@@ -228,20 +232,65 @@ async def convert_spring(
         logger.warning(msg)
         raise HTTPException(status_code=400, detail=msg)
 
-    # TODO: All other PBI 8 to integrate here
-    # This line is for quick integration when PBI 8: 3 gets merged
-    # tmp_zip_path = await initialize_springboot_zip(project_name, group_id)
+    tmp_zip_path = await initialize_springboot_zip(project_name, group_id)
+
+    with zipfile.ZipFile(tmp_zip_path, "a") as zipf:
```
MightyZanark, line 237: I think the mode would be better to be set to `zipfile.ZIP_DEFLATED` like so:
```python
zipfile.ZipFile(tmp_zip_path, "a", zipfile.ZIP_DEFLATED)
```

This is because the zip file returned by sprint initializr uses DEFLATE, so to make it the same we have to add the compression mode.

MightyZanark (changes requested):
Overall looks good! I just think that the compression level of the zip file should be updated so that it is the same as the one given from spring initializr.

app/main.py
```diff
@@ -228,20 +232,65 @@ async def convert_spring(
...
+                generate_repository_java(project_name, class_object),
+            )
+            zipf.writestr(
+                write_springboot_path(src_path, "service", class_object.get_name()),
+                generate_service_java(project_name, class_object),
+            )
+            zipf.writestr(
+                write_springboot_path(src_path, "controller", class_object.get_name()),
+                generate_springboot_controller_file(project_name, class_object),
+            )
```
DaWanAnOnli, line 285: I think this is fine for now but in the future this needs to refactored, in particular how models file is handled very differently than other files. For other files such as repository, service, and controller, the functions to render the jinja templates are located in different files. However, for models file, its render function is located in ModelsElements.

This is an architectural issue, as there will be inconsistencies across the processing of different files. This will also be a maintenance issue, as future developers will be confused as to why the models file is treated differently.

That's all from me

DaWanAnOnli (commented):
A comment regarding the handling of Springboot models file

app/main.py
NtapSlur: Thankyou for the feedback! I agree that models need to be treated like other files. In the future this is gonna be fixed!

PR #139, opened 2025-04-28
Extra set default url for swagger

NtapSlur:
- Now, / will redirect to swagger for convienience


KenKomKom:
Besides the previously broken test, I think it is a good addition for uat. I approve

PR #144, opened 2025-04-30
Bug fix spring relationship

KingOnyxx:
This pull request introduces several changes to improve the handling of relationships in Spring Boot model templates, enhance code formatting, and update corresponding test cases. The most significant updates include adding cascading behavior to relationships, refining code formatting for better readability, and aligning test cases with the updated logic.

### Enhancements to Spring Boot model templates:
* Added cascading behavior (`CascadeType.ALL`) to `@OneToOne`, `@OneToMany`, and `@ManyToMany` relationships to ensure related entities are automatically persisted or removed. (`app/models/diagram.py`: [[1]](diffhunk://#diff-78a5dbdff924e53b14761a27cb0224e46629026c314fecfdec45199126ed0680L211-R211) [[2]](diffhunk://#diff-78a5dbdff924e53b14761a27cb0224e46629026c314fecfdec45199126ed0680L245-R251) [[3]](diffhunk://#diff-78a5dbdff924e53b14761a27cb0224e46629026c314fecfdec45199126ed0680L278-R287)
* Improved formatting of annotations (e.g., added indentation for multi-line annotations) for better readability. (`app/models/diagram.py`: [[1]](diffhunk://#diff-78a5dbdff924e53b14761a27cb0224e46629026c314fecfdec45199126ed0680L245-R251) [[2]](diffhunk://#diff-78a5dbdff924e53b14761a27cb0224e46629026c314fecfdec45199126ed0680L278-R287)

### Updates to Spring Boot model templates:
* Fixed template logic for iterating over fields and relationships to ensure proper formatting. (`app/templates/springboot/model.j2`: [app/templates/springboot/model.j2L33-R43](diffhunk://#diff-6b31ef6d9dc239adcf11b4e80bc16bfb4f5e00d6f68a5738c142974efb6fbba6L33-R43))

### Test case updates:
* Updated test cases to reflect the addition of cascading behavior and improved formatting in relationship annotations. (`tests/test_diagram.py`: [[1]](diffhunk://#diff-53c4826ad5ed4307b804c033bc147f28c124d509798b5edad005021e55f0f030L386-R386) [[2]](diffhunk://#diff-53c4826ad5ed4307b804c033bc147f28c124d509798b5edad005021e55f0f030L398-R398) [[3]](diffhunk://#diff-53c4826ad5ed4307b804c033bc147f28c124d509798b5edad005021e55f0f030L411-R411) [[4]](diffhunk://#diff-53c4826ad5ed4307b804c033bc147f28c124d509798b5edad005021e55f0f030L423-R427) [[5]](diffhunk://#diff-53c4826ad5ed4307b804c033bc147f28c124d509798b5edad005021e55f0f030L453-R457)

These changes collectively enhance the maintainability and functionality of the Spring Boot model generation process.

### DIFF
--- a/app/models/diagram.py
+++ b/app/models/diagram.py
@@ -208,7 +208,7 @@ class OneToOneRelationshipObject(AbstractRelationshipObject):
             rel_type = f'@OneToOne(mappedBy="{source.replace(" ", "_")}")'
             join = None
         else:
-            rel_type = "@OneToOne"
+            rel_type = "@OneToOne(cascade = CascadeType.ALL)"
             join = f'@JoinColumn(name = "{source.replace(" ", "_")}_id")'
         var = f"private {to_pascal_case(target)} {to_camel_case(target)};"
         return {"name": var, "type": rel_type, "join": join}
@@ -242,13 +242,13 @@ class ManyToOneRelationshipObject(AbstractRelationshipObject):
         source = self.get_source_class().get_name().lower()
         target = self.get_target_class().get_name()
         if self.get_source_class_own_amount() == "1":
-            rel_type = f'@ManyToOne(mappedBy="{source.replace(" ", "_")}_id")\n'
-            rel_type += f'@JsonIgnoreProperties("{source.replace(" ", "_")}s")'
+            rel_type = f'@ManyToOne(mappedBy="{source.replace(" ", "_")}_id")'
+            rel_type += f'\n\t@JsonIgnoreProperties("{source.replace(" ", "_")}s")'
             join = None
             var = f"private {to_pascal_case(target)} {to_camel_case(target)};"
         else:
-            rel_type = "@OneToMany\n"
-            rel_type += "@JsonIgnore"
+            rel_type = "@OneToMany(cascade = CascadeType.ALL)"
+            rel_type += "\n\t@JsonIgnore"
             join = f'@JoinColumn(name = "{source.replace(" ", "_")}_id")'
             var = f"private List<{to_pascal_case(target)}> {to_camel_case(target)}s;"
         return {"name": var, "type": rel_type, "join": join}
@@ -275,16 +275,16 @@ class ManyToManyRelationshipObject(AbstractRelationshipObject):
     def to_springboot_models_template(self) -> dict[str, str]:
         source = self.get_source_class().get_name().lower()
         target = self.get_target_class().get_name()
-        rel_type = "@ManyToMany\n"
-        rel_type += "@JsonIgnore"
+        rel_type = "@ManyToMany(cascade = CascadeType.ALL)"
+        rel_type += "\n\t@JsonIgnore"
         join = "@JoinTable("
-        join += f'\n\tname = "{source.replace(" ", "_")}_{target.replace(" ", "_").lower()}",'
+        join += f'\n\t\tname = "{source.replace(" ", "_")}_{target.replace(" ", "_").lower()}",'
         join += (
-            f'\n\tjoinColumns = @JoinColumn(name = "{source.replace(" ", "_")}_id"),'
+            f'\n\t\tjoinColumns = @JoinColumn(name = "{source.replace(" ", "_")}_id"),'
         )
         join += (
-            f"\n\tinverseJoinColumns = @JoinColumn("
-            f'name = "{target.replace(" ", "_").lower()}_id")\n)'
+            f"\n\t\tinverseJoinColumns = @JoinColumn("
+            f'name = "{target.replace(" ", "_").lower()}_id")\n\t)'
         )
         var = f"private List<{to_pascal_case(target)}> listOf{to_pascal_case(target)}s;"
         return {"name": var, "type": rel_type, "join": join}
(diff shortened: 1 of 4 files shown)


app/models/diagram.py
```diff
@@ -242,13 +242,13 @@ def to_springboot_models_template(self) -> dict[str, str]:
...
-            rel_type += f'@JsonIgnoreProperties("{source.replace(" ", "_")}s")'
+            rel_type = f'@ManyToOne(mappedBy="{source.replace(" ", "_")}_id")'
+            rel_type += f'\n\t@JsonIgnoreProperties("{source.replace(" ", "_")}s")'
             join = None
             var = f"private {to_pascal_case(target)} {to_camel_case(target)};"
         else:
-            rel_type = "@OneToMany\n"
-            rel_type += "@JsonIgnore"
+            rel_type = "@OneToMany(cascade = CascadeType.ALL)"
+            rel_type += "\n\t@JsonIgnore"
```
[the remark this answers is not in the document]
MightyZanark: I agree with Gemini, I think this is worth a consideration since what we want to make is something that should have minimal change in terms of the model definition, and if using `CascadeType.ALL` can make something go wrong, users might not realize that the fault lies within it.

MightyZanark (commented):
Looks good, thanks for the bug fix! I just have some comments about mixing `\t` with regular spaces. You generated the relationship string by using `\t`, but in the Jinja template, it is using standard 4 spaces. `\t` is not necessarily always equal to 4 spaces as far as I know, so I think its better to either use `\t` or 4 spaces, but not both.

app/models/diagram.py
```diff
@@ -208,7 +208,7 @@ def to_springboot_models_template(self) -> dict[str, str]:
             rel_type = f'@OneToOne(mappedBy="{source.replace(" ", "_")}")'
             join = None
         else:
-            rel_type = "@OneToOne"
+            rel_type = "@OneToOne(cascade = CascadeType.ALL)"
             join = f'@JoinColumn(name = "{source.replace(" ", "_")}_id")'
```
DaWanAnOnli, line 212: Is there no way to do this particular formatting in the Jinja file instead of using the 'join' variable. I thought the whole point of Jinja was to avoid direct string manipulation like these. This will likely result in readability issue, which will have a n effect on maintenance issue too. please consider it

app/models/diagram.py
```diff
@@ -242,13 +242,13 @@ def to_springboot_models_template(self) -> dict[str, str]:
...
+            rel_type += f'\n\t@JsonIgnoreProperties("{source.replace(" ", "_")}s")'
             join = None
             var = f"private {to_pascal_case(target)} {to_camel_case(target)};"
         else:
-            rel_type = "@OneToMany\n"
-            rel_type += "@JsonIgnore"
+            rel_type = "@OneToMany(cascade = CascadeType.ALL)"
+            rel_type += "\n\t@JsonIgnore"
             join = f'@JoinColumn(name = "{source.replace(" ", "_")}_id")'
             var = f"private List<{to_pascal_case(target)}> {to_camel_case(target)}s;"
```
DaWanAnOnli, line 253: Is there no way to do this particular formatting in the Jinja file instead of using the 'join' variable. I thought the whole point of Jinja was to avoid direct string manipulation like these. This will likely result in readability issue, which will have a n effect on maintenance issue too. please consider it

app/models/diagram.py
```diff
@@ -275,16 +275,16 @@ def to_models_code_template(self) -> dict[str, str]:
...
+        join += f'\n\t\tname = "{source.replace(" ", "_")}_{target.replace(" ", "_").lower()}",'
         join += (
-            f'\n\tjoinColumns = @JoinColumn(name = "{source.replace(" ", "_")}_id"),'
+            f'\n\t\tjoinColumns = @JoinColumn(name = "{source.replace(" ", "_")}_id"),'
         )
         join += (
-            f"\n\tinverseJoinColumns = @JoinColumn("
-            f'name = "{target.replace(" ", "_").lower()}_id")\n)'
+            f"\n\t\tinverseJoinColumns = @JoinColumn("
+            f'name = "{target.replace(" ", "_").lower()}_id")\n\t)'
```
DaWanAnOnli, line 287: Similar comment as the previous two. Is there no way to do this particular formatting in the Jinja file instead of using the 'join' variable. I thought the whole point of Jinja was to avoid direct string manipulation like these. This will likely result in readability issue, which will have a n effect on maintenance issue too. please consider it

app/models/diagram.py
```diff
@@ -242,13 +242,13 @@ def to_springboot_models_template(self) -> dict[str, str]:
...
             join = None
             var = f"private {to_pascal_case(target)} {to_camel_case(target)};"
         else:
-            rel_type = "@OneToMany\n"
-            rel_type += "@JsonIgnore"
+            rel_type = "@OneToMany(cascade = CascadeType.ALL)"
+            rel_type += "\n\t@JsonIgnore"
             join = f'@JoinColumn(name = "{source.replace(" ", "_")}_id")'
             var = f"private List<{to_pascal_case(target)}> {to_camel_case(target)}s;"
         return {"name": var, "type": rel_type, "join": join}
```
DaWanAnOnli, line 254: Similar comment as before. Is there no way to do this particular formatting in the Jinja file instead of using the 'join' variable. I thought the whole point of Jinja was to avoid direct string manipulation like these. This will likely result in readability issue, which will have a n effect on maintenance issue too. please consider it.

DaWanAnOnli (commented):
A small comment on comments

A comment regarding the use of 'join' variable insteand of using Jinja template

PR #145, opened 2025-05-02
BUG FIX - Fix Edit Setter for Inheritance

NtapSlur:
Fix issue for edit generation that won't include the inheritance

### DIFF
--- a/app/generate_service_springboot/generate_service_springboot.py
+++ b/app/generate_service_springboot/generate_service_springboot.py
@@ -34,7 +34,12 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str) 
 
 def get_all_attributes(model: ClassObject) -> list[str]:
     attributes = []
-    for attribute in model.get_fields():
+
+    fields = model.get_fields()
+    if model.get_parent() is not None:
+        fields += model.get_parent().get_fields()
+
+    for attribute in fields:
         attribute_name = attribute.get_name()
         name = attribute_name[0].upper() + attribute_name[1:]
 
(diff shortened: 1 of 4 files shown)


tests/springboot/test_generate_service.py
```diff
@@ -20,7 +20,18 @@
...
+        field_type = TypeObject()
+        field.set_name("username")
+        field_type.set_name("string")
+        field.set_type(field_type)
+        parent.add_field(field)
+
+
         class_object = ClassObject()
+        class_object.set_parent(parent)
         class_object.set_name("Cart")
```
KenKomKom, line 35: I know this is for testing purposes only, but why is Pembeli the parent of Cart class?

DaWanAnOnli (approved):
All good frome me, as there seems to be no significant changes to the code, just small adjustments.

tests/springboot/test_generate_service.py
NtapSlur: When i was making the test, i just used the first name that came up out of my mind for the parent name so maybe that's why it's not so realistic that Cart is inherited from Pembeli. But thankyou for the feedback, gonna change it with better class name

app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -20,8 +20,26 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str)
...
     attributes = []
-    for attribute in model.get_fields():
+
+    fields = model.get_fields()
+    if model.get_parent() is not None:
+        fields += model.get_parent().get_fields()
+
+    for attribute in fields:
         attribute_name = attribute.get_name()
         name = attribute_name[0].upper() + attribute_name[1:]
```
KingOnyxx, line 44: i think you can also use this "name" variable in the else block instead of calling it again to avoid code duplication

app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -20,8 +20,26 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str)
...
+def get_all_attributes(model: ClassObject) -> list[str]:
     attributes = []
-    for attribute in model.get_fields():
+
+    fields = model.get_fields()
+    if model.get_parent() is not None:
+        fields += model.get_parent().get_fields()
+
+    for attribute in fields:
         attribute_name = attribute.get_name()
```
KingOnyxx, line 43: i think you can use this variable for the getter variable in if-block

KingOnyxx (changes requested):
Thank you for fixing bugs for edit setter! After reviewing, i found a couple of things that you can implement to improve code quality. Please consider changing them before merging. Other than that, the rest looks good to me.

app/generate_service_springboot/generate_service_springboot.py
MightyZanark, line 44: Why don't you use the `to_pascal_case` function in `app/utils.py` instead? Are there any guarantees the attributes will just be one word?

MightyZanark (changes requested):
Looks good, but there is a point in where I'd like you to change before merging.

PR #148, opened 2025-05-03
fix: duplicate application.properties and build.gradle.kts file

MightyZanark:
- Fixed duplicate files issues by writing them in Spring Initializr instead.
- Added some integration test code to make sure that Spring Initializr really writes what we wanted to the correct file. This test should not be run under normal circumstances. To ignore it, you can do `python -m pytest --ignore=tests/integration`, which will ignore all integration tests.

### DIFF
--- a/tests/integration/test_integration_initialize_spring.py
+++ b/tests/integration/test_integration_initialize_spring.py
@@ -69,6 +69,7 @@ class TestIntegrationInitializeSpring(unittest.IsolatedAsyncioTestCase):
                     "org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0",
                     buildf.read().decode("utf-8"),
                 )
+        os.remove(tmp_zip_path)
 
     async def test_application_properties_properly_filled(self):
         expected = ""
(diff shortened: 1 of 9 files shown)


tests/springboot/test_generate_project.py
```diff
@@ -100,26 +99,26 @@ def unzip_result(context):
...
-def check_repository_folder(context):
+def check_repository_folder(context: dict):
     print("aaaa", context["file_list"])
     for model in ["ShapeRepository", "CircleRepository", "CirclesRepository"]:
         assert f"{context['src_path']}/repository/{model}.java" in context["file_list"]
 
 
 @then("the zip contains service folder for all models")
-def check_service_folder(context):
+def check_service_folder(context: dict):
```
KenKomKom, line 115: Thank you for finally fixing this file, it has ruined my day more times than I can count

tests/integration/test_integration_initialize_spring.py
```diff
@@ -0,0 +1,95 @@
...
+            "run",
+            "--rm",
+            "--name",
+            "spring",
+            "-p",
+            "8080:8080",
+            "-e",
+            "HTTP_PROXY_PORT=9999",
+            "-d",
+            "mightyzanark/spring-initializr:1.0.2",
```
KenKomKom, line 36: Is it better to assign this value to the .env file for privacy sake?

MightyZanark: I don't think it matters that much since the container itself should not contain any private information, even if someone were to go dissect each layer. Anything on the image should already be available on our other repo, [spring-initializr](https://github.com/placeholder-motxt/spring-initializr), which is already open source to begin with.

KingOnyxx (approved):
Thank you for implementing these changes! After reviewing, I didn't any problem that's concerning. Approved!

PR #149, opened 2025-05-03
Extra 10 add motxt signature

KenKomKom:
On this PR, I've added motxt signature on files generated with motxt, mainly models.py, views.py, urls.py, springboot service, springboot models, controllers, repositories

Cheers!🍻

### DIFF
--- a/app/templates/repository.java.j2
+++ b/app/templates/repository.java.j2
@@ -1,3 +1,8 @@
+/*
+This code is generated using MoTxT,
+checkout more about us on https://motxt.ppl.cs.ui.ac.id
+*/
+
 package {{group_id}}.{{project_name}}.repository;
 
 import {{group_id}}.{{project_name}}.model.{{class_name}};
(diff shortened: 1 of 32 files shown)


app/templates/repository.java.j2
```diff
@@ -1,3 +1,8 @@
+/*
+This code is generated using MoTxT,
+checkout more about us on https://motxt.ppl.cs.ui.ac.id
+*/
```
MightyZanark, line 4: Rather than using 
```java
/* ...
...
*/
```
I think it would be better to use this format: 
```java
/* 
 * ... 
 */
``` 
Since it also aligns with most other regular "header" formats, like copyright for example: https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/java/lang/ProcessEnvironment.java

MightyZanark (approved):
Looks good! This way if people decided to leave these headers be in their code and make it public, we can also gain exposure. I just have a small suggestion for the Java header format.

DaWanAnOnli (approved):
All good from me, since the changes made do not change the logic of the system.

KingOnyxx (approved):
Thank you for implementing these changes! After reviewing, I didn't see any problem that's concerning. Approved!

PR #151, opened 2025-05-06
Pbi 10 1 parse return value type

DaWanAnOnli:
Finished PBI-10: 1 - Change parsing of sequence diagram to include return value type

I have edited the parse method to handle return values for method calls with the following format:

returnVarName: returnVarType

example:
book: string
count: integer

Please give feedback, especially those who handled parsing of sequence diagram in Spring 1, whether I have edited the correct parts or not.

Deadline comment: 7 May 2025 12.00 PM

By the way, the base branch is correct: EXTRA-9, as I will be changing this branch name to PBI-10 after it is graded. I did this because EXTRA-9 in Sprint 3 is pretty much the same as PBI-10 in Sprint 4.


tests/test_parse_json_to_object_seq.py
```diff
@@ -575,3 +576,106 @@ def test_invalid_param_name(self):
...
+        name, typ = parser.process_return_variable("payload: CustomType")
+        self.assertEqual(name, "payload")
+        self.assertEqual(typ.get_name(), "CustomType")  # untouched
+
+    def test_unicode_name(self):
+        """Non-ASCII variable names should parse fine."""
+        parser = ParseJsonToObjectSeq()
+        name, typ = parser.process_return_variable("имя: Boolean")
+        exp = TypeObject()
+        exp.set_name("bool")
```
KenKomKom, line 680: I think this comment is for all of us. I think we should at least dip our toes a little bit into BDD testing during PPL so that we can learn something new.

app/parse_json_to_object_seq.py
```diff
@@ -379,13 +379,33 @@ def parse(self):
...
+    def process_return_variable(self, ret_var: str) -> tuple[str, TypeObject]:
+        name_type_pair = ret_var.split(sep=":")
+        if len(name_type_pair) != 2:
+            raise ValueError(
+                f"Invalid return variable: {ret_var}!"
+                "Format must be <returnVariableName>: <returnVariableType>"
+                "\nExample: person: String"
+            )
+        ret_var_name = name_type_pair[0].strip()
+        ret_var_type_name = name_type_pair[1].strip()
```
NtapSlur, line 400: Do we need to do TYPE_MAPPING for this code?  Just to make sure that the variable type exists in Python/Java

tests/test_parse_json_to_object_seq.py
DaWanAnOnli: Okay

app/parse_json_to_object_seq.py
DaWanAnOnli: Yes. I believe I've already done that.

PR #152, opened 2025-05-07
Pbi 12 1 handle parsing aggregation

KenKomKom:
On this PR, I have implemented parsing aggregation relationship on JETUML. I also added an ENUM for relationship types which could be useful for composition relationship.
Cheers!🍻


app/parse_json_to_object_class.py
```diff
@@ -64,6 +65,16 @@ def parse_relationships(
             class_from_id = classes[edge["start"]]
             class_to_id = classes[edge["end"]]
 
+            if "type" in edge.keys() and "Aggregation Type" in edge.keys():
```
MightyZanark, line 68: Is it not enough to just check if `Aggregation Type` is one of the keys instead of having to also check if `type` is one of the keys? The if block doesn't seem to even access `type`.

MightyZanark (approved):
Overall looks good! Just have a simple question.

app/parse_class_pattern/parse_relationship_strategy.py
```diff
@@ -25,9 +27,23 @@ def create_relationship(
...
+        # It needs to be swapped because the "end" on JETUML is the class that is "a part of"
+        if ro.get_type() == RelationshipType.AGGREGATION:
+            ro.set_source_class(class_to_id)
+            ro.set_target_class(class_from_id)
+            ro.set_source_class_own_amount(edge["endLabel"])
+            ro.set_target_class_own_amount(edge["startLabel"])
+            class_to_id.add_relationship(ro)
+            return
+
+        ro.set_target_class_own_amount(edge["startLabel"])
```
KingOnyxx, line 46: why are you only implementing for django only? I think you should also add implementation for spring boot

KingOnyxx (changes requested):
Thank you for adding these changes! After reviewing, i have a question below regarding your works. Other than that, the rest looks good to me.

app/parse_class_pattern/parse_relationship_strategy.py
DaWanAnOnli: I agree. I believe this issue is relevant for both DJango and Springboot, therefore it should be implemented on both. This will affect maintenance and architecture, because it would be confusing for future developers as to why it is only implemented for Django and not Springboot.

PR #153, opened 2025-05-07
Pbi 12 5 handle aggregation in django code generation

KenKomKom:
On this PR, I've implemented a translation from aggregation relation to string on to_models_django for one to one and one to many relationships. Many to many is not affected by whether a relation is aggregation or not based on my understanding.
Cheers! 🍻


MightyZanark (approved):
Looks good! It just seem like there are duplicate commits with the other PR, which in my opinion should be avoided.

PR #155, opened 2025-05-09
EXTRA-6: Enhance Attribute Setter for Springboot for Multilevel Inheritance

NtapSlur:
Now, generated edit endpoint for Springboot can support multilevel inheritance, especially for the setter and getter.

Case Study:
Let's say that there is a class called VipPembeli that inherits from Pembeli. Then Pembeli inherits from User, so that the inheritance is like
User -> Pembeli -> VipPembeli

Before:
The setter and getter will only get attributes for Pembeli and VipPembeli

Now:
The setter and getter will get all attributes from User, Pembeli, and VipPembeli

### DIFF
--- a/app/generate_service_springboot/generate_service_springboot.py
+++ b/app/generate_service_springboot/generate_service_springboot.py
@@ -40,7 +40,12 @@ def get_all_attributes(model: ClassObject) -> list[str]:
 
     parent = model.get_parent()
 
+    accessed_parent = []
+
     while parent is not None:
+        if parent.get_name() in accessed_parent:
+            raise ValueError("Cyclic Inheritance detected! It should not be allowed!")
+        accessed_parent.append(parent.get_name())
         fields += parent.get_fields()
         parent = parent.get_parent()
 
--- a/app/generate_service_springboot/generate_service_springboot.py
+++ b/app/generate_service_springboot/generate_service_springboot.py
@@ -35,9 +35,14 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str) 
 def get_all_attributes(model: ClassObject) -> list[str]:
     attributes = []
 
-    fields = model.get_fields()
-    if model.get_parent() is not None:
-        fields += model.get_parent().get_fields()
+    fields = []
+    fields += model.get_fields()
+
+    parent = model.get_parent()
+
+    while parent is not None:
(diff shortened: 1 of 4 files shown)


app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -35,9 +35,14 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str)
...
-    if model.get_parent() is not None:
-        fields += model.get_parent().get_fields()
+    fields = []
+    fields += model.get_fields()
+
+    parent = model.get_parent()
+
+    while parent is not None:
+        fields += parent.get_fields()
+        parent = parent.get_parent()
```
KingOnyxx, line 50: i think its better to make it a list to save all parents because It could make the codebase more adaptable to models with complex relationships.  Also is there a need to handle potential cyclic inheritance scenarios? If so, should there be a safeguard or limit in the loop to avoid infinite recursion?

tests/springboot/test_generate_service.py
```diff
@@ -58,6 +58,47 @@ def test_generate_service(self):
...
+        parent.set_parent(class_object)
+        parent.set_is_public(True)
+
+        output = generate_service_java(project_name, parent, "com.example")
+
+        assert "existingPembeli.setId(pembeli.getId())" in output  # Level: AsbtractUser
+        assert "existingPembeli.setEmail(pembeli.getEmail())" in output  # Level: User
+        assert (
+            "existingPembeli.setUsername(pembeli.getUsername())" in output
+        )  # Level: Pembeli
```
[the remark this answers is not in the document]
KingOnyxx: i agree with gemini, Should there also be a negative edge test case to check how the code behaves with invalid or cyclic parent relationships?

KingOnyxx (changes requested):
Thank you for adding these changes! After reviewing, i found a couple of things that you can implement to handle edge case which is cyclic inheritance. Please consider changing them before merging. Other than that, the rest looks good to me.

app/generate_service_springboot/generate_service_springboot.py
NtapSlur: Thank you for the suggestion. If you didn't tell about cyclic inheritance, i wouldn't know to handle it. Gonna handle it!

PR #160, opened 2025-05-10
PBI-10-5: Insert Method Calls From Sequence Diagram

NtapSlur:
- SequenceDiagram parser now can handle data type for argument in method call
- Implemented method call insertion into service

P.s. This PR doesn't handle modelService call and integration with main.py. It will be done in PBI-10-7


app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -36,9 +36,19 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str)
 def get_all_attributes(model: ClassObject) -> list[str]:
     attributes = []
 
-    fields = model.get_fields()
-    if model.get_parent() is not None:
-        fields += model.get_parent().get_fields()
+    fields = []
+    fields += model.get_fields()
```
MightyZanark: Why not just make it `fields = model.get_fields()`?

app/generate_service_springboot/generate_service_springboot.py
```diff
@@ -36,9 +36,19 @@ def generate_service_java(project_name: str, model: ClassObject, group_id: str)
...
+
+    parent = model.get_parent()
+
+    accessed_parent = []
```
MightyZanark: I think using a set would be better

app/models/methods.py
```diff
@@ -187,6 +187,30 @@ def method_name(request, instance, param1, param2) -> str:
...
+        method_call_string = []
+
+        if self.__calls != []:
+            for call in self.__calls:
+                result = call.print_springboot_style_template()
+                arguments = ""
+
+                for elem in result["arguments"]:
+                    arguments += elem["argument_name"]
+                    arguments += ", "
```
MightyZanark: You can just use the join method here instead of this so that you don't need to care about the extra comma
`", ".join(elem["argument_name"] for elem in result["arguments"])`

app/parse_json_to_object_seq.py
```diff
@@ -218,8 +228,14 @@ def check_for_duplicate_attribute(
                     f"Duplicate attribute '{param}' on sequence diagram \n\
 please remove one of the parameters"
                 )
+
             param_obj = ParameterObject()
             param_obj.set_name(param)
+            if param_type != "":
+                param_object = TypeObject()
+                param_object.set_name(param_type)
```
MightyZanark: Shouldn't this be type object?

tests/springboot/test_valid_json_seq.txt
MightyZanark, line 1: Please minify this to save on space.

MightyZanark (changes requested):
Overall looks good, but there are some changes regarding readability I would like you to do.

app/generate_service_springboot/generate_service_springboot.py
NtapSlur: I remembered that Python's List is Mutable so I was afraid that if i use model.get_fields() directly, it can change the real value inside the model. To avoid that, I create a new empty list and i concat the elements

app/generate_service_springboot/generate_service_springboot.py
NtapSlur: Thank you for the suggestion. Gonna change it!

app/parse_json_to_object_seq.py
NtapSlur: Oops yes, seems the variable name is bad and not descriptive... gonna change the variable name!

app/models/methods.py
NtapSlur: Thank you for the suggestion, gonna add it!

app/generate_service_springboot/generate_service_springboot.py
MightyZanark: You can use deepcopy so that `get_fields` returns a copy as well

PR #168, opened 2025-05-17
feat: make urls.py and requirements.txt write directly to zip

MightyZanark:
This PR only have some slight changes and nothing new as it is only making the writing of `urls.py` and `requirements.txt` straight to the zip file instead of having to write it to the disk first. I have also deleted the `download_file` function as that does not seem to really be used anymore. For all the zip files I have also made it so that the contents are written in DEFLATE mode, so the files are actually compressed and not just plain text in the zip.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1,7 +1,8 @@
 import pytest
 from pydantic import ValidationError
 
-from app.main import ConvertRequest, DownloadRequest
+from app.main import ConvertRequest
+from app.model import DownloadRequest
 
 
 def test_valid_download_request():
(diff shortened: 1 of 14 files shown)


tests/test_model.py
```diff
@@ -1,7 +1,8 @@
 import pytest
```
KenKomKom, line 1: Although it is probably unnecessary to test the duplicate checker and style. I think we should still make a unit test for it. For sanity's sake!

DaWanAnOnli (approved):
Looks good to me :)

PR #172, opened 2025-05-18
Extra topo sort modelspy

NtapSlur:
Implements topological sort for models.py django so that there will be no parents that are called after the child


MightyZanark (approved):
Looks good to me! Just want to ask, this means that any ordering built in from JetUML will be disregarded and be listed correctly right (at least for Django since the models are usually in one file)?

NtapSlur:
> Looks good to me! Just want to ask, this means that any ordering built in from JetUML will be disregarded and be listed correctly right (at least for Django since the models are usually in one file)?

Yes, but the sorting will happen only for models.py (and only Django that requires the sorting). I've tested it using multi-level inheritance that is sorted upside down from JetUML and the sort works and cause no error :DDD

NtapSlur:
OOPS, having you said that, i think cyclic inheritance is not handled propertly, gonna update this now

